PR #2, opened 2019-07-18
Move from rdp proto

vaffeine:
- Updated code to the latest version from devolutions-jet/rdp-proto (including rustdoc!)
- Refactored code a little to be more independent of IronRDP and devolutions-jet


awakecoding:
Changes have been published in version 0.2.0

PR #5, opened 2019-10-15
WinAPI-like API

vaffeine:
- Changed library API to be more WinAPI-like, so people familiar with WinAPI SSPI feel more comfortable with this library
- Added wrappers around WinAPI SSPI, that can easily be used instead of a from-scratch implementation. You just need to replace a usage statement.
- Added better README, rustdoc and examples. Examples are based on the SSPI sample code from msdn.

Currently, only NTLM SSP is changed, because CredSSP turned out to be more complex than expected. The old CredSSP is now in `sspi::internal` module. But it is coming next :)


vaffeine:
Verified by QA, can be merged now

PR #7, opened 2020-08-08
add the ability to use a credentials with empty password in ntlm

puuuuh:
Empty password is allowed, for example in rdp


awakecoding:
I don't think we ever supported empty passwords in FreeRDP, is this really a legitimate feature in 2020 (RDP empty passwords?).

Since you mention RDP, you may be interested in the project for which sspi-rs was written for: https://github.com/devolutions/IronRDP

I will merge this for the sake of SSPI completeness still.

PR #8, opened 2020-09-29
Fix CredSSP version mismatch crash.

acelletti:
This PR fixes an issue with using NLA with servers that are not updated to support CredSSP v6. 

If the CredSSP version mismatches in the `TSRequest`, what happens with the old code is that we start with v6 for the first request, then as we get the reply with v5 all consecutive reply from us have v5 set (this is caused by `ts_request_version` being set to None). The other side will thread this as an error and close the connection.

By explicitly setting `ts_request_version` to a specific version we make sure that our side's version remains consistent through the exchange and avoid erroring when the versions mismatch (tested with W2008 that uses v5, it fixes the issue).

### DIFF
--- a/src/sspi/internal/credssp.rs
+++ b/src/sspi/internal/credssp.rs
@@ -29,7 +29,7 @@ use crate::{
     },
     AcceptSecurityContextResult, AcquireCredentialsHandleResult, InitializeSecurityContextResult,
 };
-use ts_request::NONCE_SIZE;
+use ts_request::{NONCE_SIZE, TS_REQUEST_VERSION};
 
 pub const EARLY_USER_AUTH_RESULT_PDU_SIZE: usize = 4;
 
@@ -169,7 +169,7 @@ pub struct CredSspClient {
     cred_ssp_mode: CredSspMode,
     client_nonce: [u8; NONCE_SIZE],
     credentials_handle: Option<AuthIdentityBuffers>,
-    ts_request_version: Option<u32>,
+    ts_request_version: u32,
 }
 
 impl CredSspClient {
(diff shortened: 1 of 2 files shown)


awakecoding:
@acelletti thanks for the fix. I am curious, are you also using IronRDP?

acelletti:
> @acelletti thanks for the fix. I am curious, are you also using IronRDP?

No, I use your credssp library only, but I am aware of the IronRDP project. It would be awesome if it eventually led to a freerdp implementation in rust, so I am keeping my fingers crossed.

awakecoding:
IronRDP is not ready for the prime time yet, but I would really like to bring it up to speed with most of the core RDP protocol. FreeRDP is very mature, but since it is written in C it doesn't provide the guarantees that Rust provides in terms of security, putting aside the fact that IronRDP + sspi-rs can be compiled to WebAssembly. I'm asking because I would definitely like to get more people involved with IronRDP to keep it growing. Since you are using sspi-rs for NLA, you must be using an RDP stack, is it a proprietary one, or a modified open source RDP stack?

acelletti:
It's a very small proprietary component we use for credential injection. 
Is there any chance of getting a new version of `sspi-rs` with this fix?

awakecoding:
I'll bump the version and publish it later today. Are you doing proxy-based credential injection? We've been working on something similar in https://github.com/devolutions/devolutions-jet using IronRDP + sspi-rs. I don't mind the fact that it is being used in a proprietary component, I'm just curious to learn more to compare our approaches.

awakecoding:
@acelletti I have just published sspi-rs 0.3.1 on crates.io. Have fun!

PR #11, opened 2022-05-09
Portable Kerberos SSPI module in Rust

RRRadicalEdward:
Hello, I've implemented the `Kerberos` protocol and added C-bindings to make this crate exportable as a dynamic C library.

`Kerberos` implementation details:

* Currently two encryption types are supported: `AES256_CTS_HMAC_SHA1_96` and `AES128_CTS_HMAC_SHA1_96`.
* `Kerberos` is integrated into the `CredSSP` too.
* Can send Kerberos messages to the `KDC server` directly or to the `KDC proxy`.
* It reaches `KDC` using the `NetworkClient` instance. This trait has two methods: `send` and `send_http`. The first one uses for sending messages to the `KDC` and the second one is for sending messages to the `KDC proxy`.
* In restricted environments where our `NetworkClient` implementation is not usable, user can pass their own implementation through `KerberosConfig`.
* In FreeRDP, where users are unable to pass the custom `KerberosConfig` structure, it looks for two environment variables: `KDC_TYPE` and `SSPI_RS_KERBEROS_URL`. The first one can have two possible values (`KDC` and `KDC_PROXY`) and the second specifies how it should reach the `KDC`. The second one contains the `Kerberos` destination address.

C-bindings implementation details:

* all C-bindings has written in separate crate `csspi` in the workspace
* We export two function tables with functions pointers: [`SecurityFunctionTableA`](https://docs.microsoft.com/en-us/windows/win32/api/sspi/ns-sspi-securityfunctiontablea) and [`SecurityFunctionTableW`](https://docs.microsoft.com/en-us/windows/win32/api/sspi/ns-sspi-securityfunctiontablew). On Windows FreeRDP uses W-table and A-table on Linux.
* The same types have different sizes on Windows and Linux. This is why I've added os-dependent macros for some structures.
* Functions that are not used for RDP (`CredSSP`) authentication return `UnsupportedFunction` error.

Useful links:

* [RFC 4120](https://datatracker.ietf.org/doc/html/rfc4120) - Kerberos RFC.
* [MS-KILE](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-KILE/%5bMS-KILE%5d.pdf) is a specification that contains all information about any Windows behavior that differs from the original Kerberos Protocol.
* [RFC 3961](https://www.rfc-editor.org/rfc/rfc3961.html) - Encryption and Checksum Specifications for Kerberos 5.
* [RFC 4121](https://datatracker.ietf.org/doc/html/rfc4121) - The Kerberos Version 5 Generic Security Service Application Program Interface (GSS-API) Mechanism.
* [SSPI](https://docs.microsoft.com/en-us/windows/win32/api/sspi/) functions list.


.gitignore
```diff
@@ -1,4 +1,5 @@
-/target
+target/
+csspi/target/
```
CBenoit, line 2: I believe that if we just specify `target` (`target/`?), it works.
[Hyper appears to be doing that](https://github.com/hyperium/hyper/blob/master/.gitignore#L1)

Cargo.toml
```diff
@@ -4,11 +4,14 @@ version = "0.3.2"
...
 homepage = "https://github.com/devolutions/sspi-rs"
 repository = "https://github.com/devolutions/sspi-rs"
-keywords = ["ntlm", "auth", "sspi", "windows"]
+keywords = ["ntlm", "auth", "sspi", "windows", "kerberos"]
 authors = ["Vlad Aleksashyn <vlad.al.dp@gmail.com>"]
 description = "A Rust implementation of the Security Support Provider Interface (SSPI) API"
 edition = "2018"
 
+[features]
+with_reqwest_network_client = ["reqwest"]
```
CBenoit, line 13: Simply
```suggestion
network_client = ["reqwest"]
```

src/sspi/kerberos.rs
```diff
@@ -0,0 +1,598 @@
...
+            generate_authenticator_for_tgs_ap_req, generate_neg_ap_req, generate_neg_token_init,
+            generate_tgs_req,
+        },
+        AES128_CTS_HMAC_SHA1_96, AES256_CTS_HMAC_SHA1_96,
+    },
+    config::{KdcType, KerberosConfig},
+    encryption_params::EncryptionParams,
+    server::extractors::extract_tgt_ticket,
+    utils::{serialize_message, utf16_bytes_to_utf8_string},
+};
```
CBenoit, line 37: Could you check [IronRDP’s rustfmt.toml](https://github.com/Devolutions/IronRDP/blob/master/rustfmt.toml) and run `cargo +nightly fmt` with the same configuration, please?

I think the only issue with `group_imports = "StdExternalCrate"` is the way it happily mixes `pub use`s with regular `use`s

CBenoit (commented):
Thank you for fixing clippy lints!

Besides the other points, could you also organize the csspi package similar to how it’s done in picky?
https://github.com/Devolutions/picky-rs/tree/master/ffi

Thank you!

.gitignore
RRRadicalEdward: Done

Cargo.toml
RRRadicalEdward: Done

Cargo.toml
```diff
@@ -26,6 +29,16 @@ lazy_static = "1.2"
...
 serde_derive = "1.0"
 winapi = { version = "0.3", features = ["sspi", "rpcdce", "impl-default", "timezoneapi", "wincrypt"] }
+url = "2.2.2"
+reqwest = { version = "0.11", features = ["blocking"], optional = true }
+picky-krb = { git = "https://github.com/Devolutions/devolutions-gateway.git", branch = "portable-kerberos" }
+picky-asn1 = { version = "0.5.0", features = ["chrono_conversion"] }
+picky-asn1-der = { git = "https://github.com/Devolutions/picky-rs.git", branch = "application-tag-improvements-and-nla-oids" }
+picky-asn1-x509 = { git = "https://github.com/Devolutions/picky-rs.git", branch = "application-tag-improvements-and-nla-oids" }
+kerberos_crypto = "0.3.6"
+kerberos_constants = "0.0.9"
```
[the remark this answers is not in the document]
awakecoding: @RRRadicalEdward we can merge with the dependency, but we should look into importing the required code bits directly in sspi-rs afterward to maintain it on our side.

csspi/src/w.rs
```diff
@@ -0,0 +1,371 @@
...
+
+#[repr(C)]
+pub struct SecWinntAuthIdentityW {
+    user: *const c_ushort,
+    user_length: c_ulong,
+    domain: *const c_ushort,
+    domain_length: c_ulong,
+    password: *const c_ushort,
+    password_length: c_ulong,
+    flags: c_ulong,
```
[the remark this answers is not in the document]
RRRadicalEdward: I made it public

src/sspi/kerberos.rs
RRRadicalEdward: I ran `cargo fmt`. Also, I put the rustfmt.toml from `IronRDP` into the project.

CBenoit: Perfect!

CBenoit:
> I improved the organization of the `csspi`, but I'm not sure if it's exactly what you wanted.

Sorry, I think I wasn’t specific enough. If it’s okay for you, I’ll just clean the last bits before merging (update to published dependencies and small details related to workspace). It looks to me otherwise!

PR #17, opened 2022-05-26
Fix target name handling in InitializeSecurityContextA

awakecoding:


awakecoding:
@RRRadicalEdward just so you, it was missing the code required in the "A" function to pass the target name properly ;)

thenextman (approved):
LGTM

PR #20, opened 2022-05-27
Fix Kerberos name type and realm

awakecoding:
Tested with multiple combinations of username and domain information, it makes UPN + empty domain work.


awakecoding:
@RRRadicalEdward FYI, some fixes on my side to handle various username and domain combinations. It works a lot better now for me.

CBenoit (approved):
LGTM!

PR #21, opened 2022-05-27
Simplify KDC configuration to use SSPI_KDC_URL environment variable

awakecoding:
Simplify KDC configuration detection to use a single SSPI_KDC_URL environment variable instead of KDC_TYPE + SSPI_RS_KERBEROS_URL. The KDC type is simply detected from the KDC URL protocol scheme (http/https = KDC proxy). If no scheme is provided in the URL, TCP is assumed instead of failing. I also noticed that HTTPS validation would fail for me on Linux even after I correctly imported my private trusted root CA, it turns out we had to enable the "rustls-tls-native-roots" feature in reqwest to make it use the system trusted root CAs.

Here are sample SSPI_KDC_URL values that work for me now:
- IT-HELP-DC.ad.it-help.ninja
- IT-HELP-DC.ad.it-help.ninja:88
- tcp://IT-HELP-DC.ad.it-help.ninja:88
- https://kdc.ad.it-help.ninja:4343/KdcProxy

We currently still panic if the KDC configuration cannot be found while creating the KerberosConfig object, we'll have to handle that more gracefully and start building the code required to do proper KDC server detection as the connection is made.


awakecoding:
@RRRadicalEdward FYI, this changes the environment variables required. It only requires the KDC URL now and it's more permissive with regards to the URL format.

CBenoit (approved):
LGTM

RRRadicalEdward:
@awakecoding Hello. 
Could you check if you have the HTTPS validation issue with the "rustls-tls-native-roots" feature of reqwest disabled on Windows?
Enabling the "rustls-tls-native-roots" feature breaks building on Windows. I could disable the feature for Windows if there's no HTTPS validation issue without the feature.

awakecoding:
> @awakecoding Hello. 
> Could you check if you have the HTTPS validation issue with the "rustls-tls-native-roots" feature of reqwest disabled on Windows?
> Enabling the "rustls-tls-native-roots" feature makes the building of csspi fail on Windows. I could disable the feature for Windows if there's no HTTPS validation issue without the feature.

Argh, I was afraid it would cause this kind of problem on Windows. It was required for HTTPS validation on Linux, but if we're lucky it's not required for Windows. Is there a way to turn it off for Windows only?

CBenoit:
I don't think we can do that at the [`Cargo.toml` level](https://github.com/rust-lang/cargo/issues/1197#issuecomment-231795547).
However, we could add a feature and turn it on / off automatically in a build.rs au simply invoke `cargo build --features` appropriately. Do you want me to link into this?

RRRadicalEdward:
I like the idea of `build.rs`. I
I'm not sure, but there might be a possibility of turning off the linking of `Secur32.dll` from `winapi` crate directly in `build.rs`. So we could leave the `rustls-tls-native-roots` feature.  
I will see if it's possible.

awakecoding:
@RRRadicalEdward enabling this specific feature on non-Windows using a build.rs script would work. We'll have to see if certificate validation works with the KDC proxy on Windows without it

awakecoding:
@RRRadicalEdward I have another workaround, but we should really test to see that nothing breaks. schannel-rs is the crate linking to secur32.dll, and it's imported by the rustls-tls-native-roots feature, but I don't think it calls the conflicting functions we also export on our side. What I fear the most is that with /FORCE:MULTIPLE you don't really know which symbol gets exported in the end. 
image

RRRadicalEdward:
@awakecoding Do we care about KDC Proxy using a trusted certificate for HTTPS or not?
If we aren't, we could just use [danger_accept_invalid_certs](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.danger_accept_invalid_certs) 
 in `reqwest`. It forces `rustls` to accept any certificates. We use `rustls` in such a way in `IronRDP` -  https://github.com/Devolutions/IronRDP/blob/master/ironrdp_client/src/main.rs#L23.

awakecoding:
@RRRadicalEdward yes we do care about validating the KDC proxy HTTPS certificate, or at least enforce validation by default. /FORCE:MULTIPLE works for now at least, but I think I've got a better solution for Windows symbol exporting.

PR #27, opened 2022-06-16
Negotiate module

RRRadicalEdward:
I've implemented the `Negotiate` module that will choose in the runtime (during authorization) what NLA protocol to use: Kerberos or NTLM.
Kerberos is preferred but not always possible. it falls back to the NTLM In two cases:
* When the user doesn't provide explicitly KDC URL and when we can not detect KDC using DNS
* When Kerberos returns the `SEC_E_NO_CREDENTIALS` error

It detects the KDC server host If the user tries to log in with FQDN then we'll cut the domain part from it and try to find `_kerberos._tcp.<domain>`/`_kerberos._udp.<domain>` DNS SRV records. If it is found, we'll take the target host from this record.

For example: if the user tries to log in with `user@example.com` then we'll cut `example.com` from it and try to find `_kerberos._tcp.example.com`/`_kerberos._udp.example.com` DNS SRV records.

Also, I've added support for different security packages in C-bindings (`sspi-c-api`).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,6 +44,7 @@ kerberos_crypto = "0.3.6"
 kerberos_constants = "0.0.9"
 oid = "0.2.1"
 sys-info = "0.9"
+trust-dns-resolver = "0.21.2"
 
 [dev-dependencies]
 whoami = "0.5"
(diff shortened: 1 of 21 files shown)


awakecoding (approved):
LGTM

PR #32, opened 2022-07-07
C# NegotiateStream fixes

RRRadicalEdward:
C# _NegotiateStream_ calls _SSPI_ to make an authentication. If  _NegotiateStream_ uses sspi-rs as the _SSPI_ module, authentication will fail because of some issues and not implemented things in sspi-rs. This PR fixes it.

Improvements:
- Added an implementation of the _Negotiate_ structure for a server (`accept_security_context_impl` function).
- Added an explicit calling to the complete_auth_token if there wasn't one before.  To fully complete _NTLM_ authentication, the user needs to call the `complete_auth_token` function, but _NegotiateStream_ doesn't call it. 
- Added support for _SECPKG_ATTR_NEGOTIATION_INFO_ attribute in _QueryContextAttributesA/W_. _NegotiateStream_ needs it.
- Fixed context attributes. So far _NTLM_ and other implemented security packages didn't care about context attributes but _NegotiateStream_ does. I just make them in C-API as a user wants to: 
`*pf_context_attr = f_context_req;`

### DIFF
--- a/c-api/src/utils.rs
+++ b/c-api/src/utils.rs
@@ -60,27 +60,3 @@ pub unsafe fn transform_credentials_handle<'a>(
         ))
     }
 }
-
-#[macro_export]
-macro_rules! try_execute {
-    ($x:expr) => {{
-        match $x {
-            Ok(value) => value,
-            Err(err) => {
-                return err.error_type.to_u32().unwrap();
-            }
-        }
-    }};
-}
-
-#[macro_export]
-macro_rules! check_null {
-    ($x:expr) => {{
-        use num_traits::ToPrimitive;
-        use sspi::ErrorKind;
-
-        if $x.is_null() {
-            return ErrorKind::InvalidParameter.to_u32().unwrap();
-        }
-    }};
-}
--- a/c-api/src/utils.rs
+++ b/c-api/src/utils.rs
@@ -72,3 +72,15 @@ macro_rules! try_execute {
         }
     }};
 }
+
+#[macro_export]
+macro_rules! check_null {
+    ($x:expr) => {{
+        use num_traits::ToPrimitive;
+        use sspi::ErrorKind;
+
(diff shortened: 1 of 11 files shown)


src/sspi/ntlm.rs
```diff
@@ -257,6 +257,10 @@ impl SspiImpl for Ntlm {
             NtlmState::Authenticate => {
                 let input_token = SecurityBuffer::find_buffer(input, SecurityBufferType::Token)?;
 
+                if let Some(identity) = builder.credentials_handle {
+                    self.identity = identity.clone();
+                }
```
CBenoit, line 262: Equivalent to:
```suggestion
                self.identity = identity.clone();
```

c-api/src/utils.rs
```diff
@@ -72,3 +72,15 @@ macro_rules! try_execute {
         }
     }};
 }
+
+#[macro_export]
```
CBenoit, line 76: nit: `#[macro_export]` is not intended for internal macro helpers (if you remove it, I recommend putting all the helper macros in a `macros` module that must be declared before all the others, generally speaking macros must be declared before they are used).

CBenoit (approved):
Nice thank you! Just a few comments, but otherwise LGTM

src/sspi/ntlm.rs
RRRadicalEdward: It's not actually possible. `self.identity` is `Option<AuthIdentityBuffer>` and `builder.credentials_handle` is `Option<&mut Option<AuthIdentityBuffer>>`.
I changed it to `self.identity = builder.credentials_handle.and_then(|creds| creds.clone());`. I hope It's suitable for you.

CBenoit: Oh okay, in that case what you need is the `.cloned()` method. Not a big deal anyway, I'll merge

PR #33, opened 2022-07-18
 Kerberos "Change password" protocol

RRRadicalEdward:
I implemented the [Kerberos Change Password Protocol](https://datatracker.ietf.org/doc/html/rfc3244). [`ChangeAccountPasswordA/W`](https://docs.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-changeaccountpasswordw) function is used to invoke the change password process. 

Password changing works only for the _Kerberos_ security package. The library will return an error if a user tries to change the password using _NTLM_.

Also, I refactored some of the _Kerberos_ code: I replaced many similar functions (like `generate_as_req_without_pre_auth`, `generate_as_req`) with one general function and generation options as a parameter.

Docs & reference:
* [ChangeAccountPasswordW function](https://docs.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-changeaccountpasswordw)
* [RFC 3244](https://datatracker.ietf.org/doc/html/rfc3244)
* [RFC 4120](https://datatracker.ietf.org/doc/html/rfc4120)


RRRadicalEdward:
This PR should be merged after merging https://github.com/Devolutions/picky-rs/pull/155

Cargo.toml
```diff
@@ -37,9 +37,9 @@ serde_derive = "1.0"
 winapi = { version = "0.3", features = ["sspi", "rpcdce", "impl-default", "timezoneapi", "wincrypt"] }
 url = "2.2.2"
 reqwest = { version = "0.11", features = ["blocking", "rustls-tls", "rustls-tls-native-roots"], optional = true, default-features = false }
-picky-krb = "0.2.0"
-picky-asn1 = { version = "0.5.0", features = ["chrono_conversion"] }
-picky-asn1-der = "0.3.1"
+picky-krb = { git = "https://github.com/Devolutions/picky-rs.git", rev = "79e8186c798876c77e67cdcdb02a1c6f58030c6e" }
+picky-asn1 = { git = "https://github.com/Devolutions/picky-rs.git", rev = "79e8186c798876c77e67cdcdb02a1c6f58030c6e" , features = ["chrono_conversion"] }
+picky-asn1-der = { git = "https://github.com/Devolutions/picky-rs.git", rev = "79e8186c798876c77e67cdcdb02a1c6f58030c6e" }
```
RRRadicalEdward, line 42: I will update these dependencies once https://github.com/Devolutions/picky-rs/pull/155 is merged and the `picky-krb` version will be bumped

RRRadicalEdward: I updated it

CBenoit: Thank you!

c-api/src/sec_handle.rs
```diff
@@ -700,19 +700,61 @@ pub type SetCredentialsAttributesFnW =
...
+    check_null!(psz_old_password);
+    check_null!(psz_new_password);
+    check_null!(p_output);
+
+    let security_package_name = CStr::from_ptr(psz_package_name).to_str().unwrap();
+
+    let domain = CStr::from_ptr(psz_domain_name).to_str().unwrap();
+    let username = CStr::from_ptr(psz_account_name).to_str().unwrap();
+    let password = CStr::from_ptr(psz_old_password).to_str().unwrap();
+    let new_password = CStr::from_ptr(psz_new_password).to_str().unwrap();
```
CBenoit, line 725: I think it’s best to check the error instead of panicking here. I assume that most of the time the input will be ASCII and therefore valid UTF-8, but just to be safe. Unwinding a panic across FFI boundaries is pretty bad, if it were to happen.
You can probably use a macro like we used to have in picky c-api to make that more ergonomic: https://github.com/Devolutions/picky-rs/blob/eea7fa2d89dea68ca70b508933dfd75617375e34/c-api/src/helper.rs#L18
Make me think that you could maybe have a look at the `ffi_helpers` crate as well: https://github.com/Michael-F-Bryan/ffi_helpers

src/sspi/builders/change_password.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn with_old_password(&self, old_password: String) -> &Self {
+        self.inner.borrow_mut().old_password = Some(old_password);
+        self
+    }
+
+    /// Required
+    pub fn with_new_password(&self, new_password: String) -> &Self {
+        self.inner.borrow_mut().new_password = Some(new_password);
+        self
+    }
```
CBenoit, line 58: You could take a `impl Into<String>` instead of a concrete `String`. This is a bit more ergonomic on the call site 🙂

CBenoit (commented):
LGTM, but make sure there is no `unwrap`s easy to trigger at the FFI boundary. It’s (sadly) hard to completely prevent, but let’s give it our best effort😄

c-api/src/sec_handle.rs
RRRadicalEdward: I got rid of the `unwrap`s. Also, I added a panic catch in all the exported functions via the `check_panic!` macro(similar to one from `ffi_helpers`).  
It should be safer now 🙂

CBenoit (approved):
Looks good! Thank you :smile:

PR #35, opened 2022-07-28
SECBUFFER_CHANNEL_BINDINGS support

RRRadicalEdward:
I added support of the security channel bindings. Basically, the channel bindings are an md5 checksum of the token that is used to prevent MITM attacks.

1. In the case of the NTLM, this checksum is transferred as special AvPair (AvPair::ChannelBindings).
2. In the case of the Kerberos, it is transferred as a part of the checksum in the authenticator during the AP exchange. 

The token for md5 checksum calculation passed into our library as the SecurityBuffer with the type SECBUFFER_CHANNEL_BINDINGS.

References:

- https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/extended-protection-for-authentication-overview
- https://docs.microsoft.com/en-us/windows/win32/api/sspi/ns-sspi-sec_channel_bindings
- https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-NLMP/%5bMS-NLMP%5d.pdf
- https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-KILE/[MS-KILE].pdf
- https://datatracker.ietf.org/doc/html/rfc4121#section-4.1.1

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ serde_derive = "1.0"
 winapi = { version = "0.3", features = ["sspi", "rpcdce", "impl-default", "timezoneapi", "wincrypt"] }
 url = "2.2.2"
 reqwest = { version = "0.11", features = ["blocking", "rustls-tls", "rustls-tls-native-roots"], optional = true, default-features = false }
-picky-krb = "0.3.0"
+picky-krb = { path = "../picky-rs/picky-krb" }
 picky-asn1 = { version = "0.5.0", features = ["chrono_conversion"] }
 picky-asn1-der = "0.3.1"
 picky-asn1-x509 = "0.7.0"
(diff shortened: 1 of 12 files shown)


Cargo.toml
```diff
@@ -37,7 +37,7 @@ serde_derive = "1.0"
 winapi = { version = "0.3", features = ["sspi", "rpcdce", "impl-default", "timezoneapi", "wincrypt"] }
 url = "2.2.2"
 reqwest = { version = "0.11", features = ["blocking", "rustls-tls", "rustls-tls-native-roots"], optional = true, default-features = false }
-picky-krb = "0.3.0"
+picky-krb = { git = "https://github.com/Devolutions/picky-rs.git", rev = "8a05c988832722af2628d2270c46cb6b47c1c41d" }
```
RRRadicalEdward, line 40: I will update the `picky-krb` dependency once the version is bumped

awakecoding:
@CBenoit all good with me, did you want to publish picky-krb first?

CBenoit:
> @CBenoit all good with me, did you want to publish picky-krb first?

Yes, I’ll publish tomorrow

Cargo.toml
CBenoit: I just published now :slightly_smiling_face:

RRRadicalEdward: Thanks! 
I pushed a commit that sets `picky-krb` to 0.3.1

CBenoit (approved):
I took some time to check reference docs. Looks good to me!

PR #37, opened 2022-08-11
fix: structs not implementing Send trait

CBenoit:
cc @RRRadicalEdward 

Introduction of a field with the type `Box<dyn NetworkClient>` caused all wrapping structs to be non-`Send`ables. This is breaking and probably not desirable (this notably breaks Devolutions Gateway code).

Other breaking changes were introduced in version starting 0.3.3 and up to 0.3.6, the most obvious one being the `CredSspClient::new` function which now has 5 parameters instead of 3.

For this reason, I yanked all these versions and this PR bumps the version to 0.4.0.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sspi"
-version = "0.3.6"
+version = "0.4.0"
 license = "MIT/Apache-2.0"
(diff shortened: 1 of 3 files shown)


thenextman (approved):
LGTM

PR #39, opened 2022-08-15
Initial CI build

awakecoding:
Please squash. Builds on all platforms except Windows ARM64. iOS builds only a static library which we can't use, and Rust doesn't yet support dynamic linking on iOS.


thenextman (approved):
LGTM

richerlariviere (approved):
nuget.yml github workflow LGTM.

PR #43, opened 2022-09-08
Fix InitializeSecurityContext input buffer null check

awakecoding:


thenextman (approved):
LGTM

awakecoding:
@RRRadicalEdward FYI - InitializeSecurityContext usually has a null input buffer for the first call, so I'm not sure how this change was tested. I was getting invalid parameter errors until I realized there was a null pointer check for the input buffer when it's perfectly normal. By the way, I've just tried it with a user member of the Protected Users group, and it works.

PR #45, opened 2022-09-09
replace sys_info::hostname by whoami::hostname

awakecoding:
sys_info::hostname() calls hostname.exe on Windows, while whoami::hostname() calls the Win32 APIs. We had some calls to whoami::hostname() in some places already, and sys_info::hostname() was the only reason to import the sys_info crate.


thenextman (approved):
LGTM

PR #48, opened 2022-09-20
sspi-rs: update native library packaging in nuget

thenextman:
- Use `lipo` to create a universal binary for macOS. Using `osx-universal` as the RID but there doesn't seem to be an official "universal" RID currently
- The nuget packages references the native library on macOS using `NativeReference`; and only deploys the thin binaries for the PowerShell target
- Make a naive check for Android target in the .targets file to only deploy Android binaries on that platform


awakecoding (approved):
LGTM

richerlariviere (approved):

PR #61, opened 2022-11-01
PKU2U protocol implementation

RRRadicalEdward:
This PR adds an implementation of the PKU2U authorization protocol in the `sspi-rs` and updated FFI-bindings to support this protocol.
Also, I've improved the Negotiate module: now if a user will try to log in under the `AzureAD` domain then it automatically will switch to the PKU2U. 

Important note: for the PKU2U Azure AD certificate is needed with the corresponding private key. The current implementation will try to extract them from the Windows Certificate Store using WinAPI bindings. Requirements for the certificate:

* It has to be installed in the current user certificate store in the personal folder
* It has to have the corresponding private key that is marked as exportable;
* An issuer CN has to start with the `MS-Organization-P2P-Access` like `MS-Organization-P2P-Access [2021]` or `MS-Organization-P2P-Access [2022]`;
* `ClientAuth` extended key usage has to be present.

Used Docs & References:

* [PKINIT](https://www.rfc-editor.org/rfc/rfc4556)
* [NEGOEX](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-negoex/77c795cf-e522-4678-b0f1-2063c5c0561c)
* [PKU2U draft](https://datatracker.ietf.org/doc/html/draft-zhu-pku2u-09)
* [MS-KILE](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-kile/2a32282e-dd48-4ad9-a542-609804b02cc9)
* [Encryption and Checksum Specifications for Kerberos 5](https://www.rfc-editor.org/rfc/rfc3961.html)


Cargo.toml
```diff
@@ -36,19 +37,22 @@ serde = "1.0"
 serde_derive = "1.0"
 url = "2.2.2"
 reqwest = { version = "0.11", features = ["blocking", "rustls-tls", "rustls-tls-native-roots"], optional = true, default-features = false }
-picky-krb = "0.4.0"
-picky-asn1 = { version = "0.5.0", features = ["chrono_conversion"] }
-picky-asn1-der = "0.3.1"
-picky-asn1-x509 = "0.7.0"
+picky-krb = { git = "https://github.com/Devolutions/picky-rs.git", rev = "50dfc40" }
```
RRRadicalEdward, line 40: I will update the `picky-*` dependencies once https://github.com/Devolutions/picky-rs/pull/186 is merged and the new versions of the crates are published

src/sspi/negotiate.rs
```diff
@@ -69,88 +91,113 @@ pub struct Negotiate {
...
-        if let Some(filtered_protocol) = Self::filter_protocol(&self.protocol, &self.package_list) {
+        if let Some(filtered_protocol) = Self::filter_protocol(&self.protocol, &self.package_list)? {
             self.protocol = filtered_protocol;
         }
 
         Ok(())
     }
 
-    fn get_package_list_config(package_list: &Option<String>) -> (bool, bool) {
+    fn get_package_list_config(package_list: &Option<String>) -> (bool, bool, bool) {
```
CBenoit, line 140: Nitpick. It’s hard for me to understand the return value of this function without looking inside. I think it would help to return a struct with named fields instead. What do you think?

Something in the lines of:

```rust
struct PackageListConfig {
    ntlm_package: bool,
    kerberos_package: bool,
    pku2u_package: bool,
}
```

It’s not a blocker since this is internal API.

CBenoit, line 140: Make me think: what about renaming to `parse_package_list_config`? Because it appears to be parsing a string

src/sspi/pku2u/cert_utils/validation.rs
```diff
@@ -0,0 +1,50 @@
+use picky_asn1_x509::signed_data::{CertificateChoices, SignedData};
+use picky_asn1_x509::{Certificate, PublicKey};
+use rsa::{BigUint, RsaPublicKey};
```
CBenoit, line 3: Do you think we could use `picky` instead of depending on `rsa` directly? If I recall correctly, the logic in `picky::key::PublicKey` is roughly the same.
As a bonus, once `picky` gains support for elliptic curves, we get it in sspi-rs as well for free 🙂

src/sspi/pku2u.rs
```diff
@@ -0,0 +1,797 @@
...
+        Ok(SecurityStatus::Ok)
+    }
+
+    fn encrypt_message(
+        &mut self,
+        _flags: EncryptionFlags,
+        message: &mut [SecurityBuffer],
+        sequence_number: u32,
+    ) -> Result<SecurityStatus> {
+        SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Token)?;
```
CBenoit, line 215: The return value is not used so I suppose this is a sanity check making sure a "Token" buffer is in the message, am I right? I think it would help to have a comment explaining why this is necessary. Also, it’s not clear to me why the `_mut` variant is used.

src/sspi/pku2u/cert_utils/win_extraction.rs
```diff
@@ -0,0 +1,329 @@
...
+        ErrorKind::InternalError,
+        "Cannot find appropriate device certificate".into(),
+    ))
+}
+
+pub fn extract_client_p2p_cert_and_key() -> Result<(Certificate, RsaPrivateKey)> {
+    unsafe {
+        // "My\0" encoded as a wide string.
+        // More info: https://docs.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certopenstore#remarks
+        let my: [u16; 3] = [77, 121, 0];
```
CBenoit, line 251: I believe there are other system store names besides `My` such as `Root`, `TrustedPeople`, `TrustedPublisher`…
The link is not explaining the rationale behind choosing specifically `My` here. Are we sure we never need to look into other stores?

CBenoit (commented):
Good work! 👍

src/sspi/negotiate.rs
RRRadicalEdward: yes, a struct with protocol names is much better than a tuple with three bools.

RRRadicalEdward: Good point. Done

src/sspi/pku2u/cert_utils/validation.rs
RRRadicalEdward: I just forgot that we have signature generation and verification implemented in picky. 
Done.

src/sspi/pku2u/cert_utils/win_extraction.rs
RRRadicalEdward: I've noticed that a `P2P` certificate is always a personal one. I had the same in the further dev-testing. I haven't found information that P2P certificates are always there. According to what I've had, I believe it's always there. 
 Through the `My` certificate store we can access personal certificates. So It's why it's in use

src/sspi/pku2u.rs
RRRadicalEdward: yes, it's just for checking that the Token buffer is just present. I added the comment about it and changed the function used to the not `_mut` variant

src/sspi/pku2u/cert_utils/win_extraction.rs
CBenoit: I understand. Could you just add this information in a comment above? It’s really just to let a future reader know that the question has been raised and it’s not arbitrary :slightly_smiling_face:

RRRadicalEdward: Done

Cargo.toml
CBenoit: picky crates are now published on crates.io 🙂

RRRadicalEdward: @CBenoit Thanks. Now, this PR is ready to be merged 🙂

CBenoit (approved):
Thank you Alex! 🙂

PR #66, opened 2022-11-10
PKU2U: RSA signature generation fix

RRRadicalEdward:
In https://github.com/Devolutions/sspi-rs/pull/61, one of the comments was to use `picky-rs` instead of `rsa` directly for an RSA signature generation.  The changes have been made, but `picky-rs` generates hash itself and so we shouldn't generate it. Now, we generate a hash in `ssp-rs` before the sign, then, pass the generated hash to SignatureAlgorithm::RsaPkcs1v15::sign which performs hashing again. That is not correct. This PR fixes it.
Also, I fixed `clippy` warnings.


CBenoit (approved):
Thank you!

PR #69, opened 2022-11-14
chore(sspi-rs): version bump: 0.5.1

RRRadicalEdward:


RRRadicalEdward:
@CBenoit Could you also publish the new version after merging? Thank you!

PR #75, opened 2022-11-17
CI build and checks, refactor workflows

thenextman:
The `build-native` and `build-universal` steps of nuget.yml are broken out into a new workflow (build-native.yml). I added an extra matrix element for "build type" - release and debug (with the debug build enabling the `debug_mode` feature).

There is no debug nuget package produced - the debug artifacts are simply too large. However the libraries are all available as artifacts for debugging and dev.

I added a CI workflow to be called from PRs on `master`. It has the same basic steps as IronRDP (formatting, lints, tests) as well as calling the `build-native` workflow. We can validate native build on all platforms at integration time.

Additionally, I disabled failing tests (including one doc test), refactored suggestions by clippy and fixed deprecation warnings in the workflow(s).

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -1,0 +1,340 @@
+name: Build Native
+
+on:
+  workflow_call:
+
+jobs:
+  preflight:
+    name: Preflight
+    runs-on: ubuntu-20.04
+    outputs:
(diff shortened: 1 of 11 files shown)


awakecoding (approved):

CBenoit (approved):
LGTM 👍

PR #76, opened 2022-11-17
add output buffer memory allocation

awakecoding:
Add support for output buffer memory allocation in InitializeSecurityContext when the ISC_REQ_ALLOCATE_MEMORY flag is passed. The buffer has to be freed using FreeContextBuffer, and while we already have such a function, the drop_in_place() call likely didn't do anything on a void pointer. This means we've been leaking memory, so I've just commented out the call to free memory inside FreeContextBuffer, and we'll need to review buffer memory allocation/release in the SSPI API to properly handle all cases, and avoid freeing memory incorrectly.

### DIFF
--- a/ffi/src/common.rs
+++ b/ffi/src/common.rs
@@ -82,7 +82,7 @@ pub unsafe extern "system" fn AcceptSecurityContext(
             .with_output(&mut output_tokens)
             .execute();
 
-        copy_to_c_sec_buffer((*p_output).p_buffers, &output_tokens);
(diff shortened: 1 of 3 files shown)


awakecoding:
@RRRadicalEdward FYI

awakecoding:
we can wait until https://github.com/Devolutions/sspi-rs/pull/75 is merged for this PR @thenextman

CBenoit (approved):
LGTM
Richard’s PR is merged, you can rebase on master to get the checks

PR #79, opened 2022-11-18
Fix Kerberos EncryptMessage

awakecoding:
WSMan/WinRM calls EncryptMessage/DecryptMessage more times than CredSSP does, so we've hit a condition where the internal Kerberos state was incorrectly rejecting the calls.


awakecoding:
@RRRadicalEdward FYI

thenextman (approved):
LGTM

PR #87, opened 2022-12-01
Fix memory leaks 

RRRadicalEdward:
I added usage of Box::from_raw(ptr) to properly deallocate memory instead of using `drop_in_place`. I made that if a struct(such as `PackageInfo`) contains an additional pointer to data, we allocate the data right after the struct. So it can be more easily deallocated after 
Also, I wrote a simple executable that loads the SSPI library and then calls different methods from it. I used it with `cargo-valgrind` to check if all memory was deallocated properly.

### DIFF
--- a/tools/sspi-ffi-attacker/src/attack.rs
+++ b/tools/sspi-ffi-attacker/src/attack.rs
@@ -1,12 +1,13 @@
 use std::ffi::CStr;
-use std::ptr::null;
+use std::ptr::{null, null_mut};
 
 use libc::c_void;
 
 use crate::types::{
     InitSecurityInterfaceA, InitSecurityInterfaceW, PSecPkgInfoA, PSecPkgInfoW, PSecurityFunctionTableA,
-    PSecurityFunctionTableW, SecBuffer, SecBufferDesc, SecHandle, SecWinntAuthIdentityA, SecWinntAuthIdentityW,
-    SspiEncodeStringsAsAuthIdentityFn, SspiFreeAuthIdentityFn, SECURITY_INTEGER, SEC_CHAR, SEC_WCHAR,
+    PSecurityFunctionTableW, SecBuffer, SecBufferDesc, SecHandle, SecPkgInfoA, SecPkgInfoW, SecWinntAuthIdentityA,
+    SecWinntAuthIdentityW, SspiEncodeStringsAsAuthIdentityFn, SspiFreeAuthIdentityFn, SECURITY_INTEGER, SEC_CHAR,
+    SEC_WCHAR,
 };
 use crate::utils::{c_w_str_to_string, get_library_fn, load_library};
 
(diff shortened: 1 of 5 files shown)


ffi/src/common.rs
```diff
@@ -218,8 +222,9 @@ pub type VerifySignatureFn = extern "system" fn(PCtxtHandle, PSecBufferDesc, c_u
 #[cfg_attr(feature = "debug_mode", instrument(skip_all))]
 #[cfg_attr(windows, rename_symbol(to = "Rust_FreeContextBuffer"))]
 #[no_mangle]
-pub unsafe extern "system" fn FreeContextBuffer(_pv_context_buffer: *mut c_void) -> SecurityStatus {
-    // FIXME: patch leak
+pub unsafe extern "system" fn FreeContextBuffer(pv_context_buffer: *mut c_void) -> SecurityStatus {
+    let _ = Box::from_raw(pv_context_buffer as *mut _);
```
CBenoit, line 226: It’s unclear to me how the type is inferred at the `_` here :thinking: 
I would think it changes nothing (it gives `*mut c_void`).
What kind of data is that supposed to be? `SecBuffer`?

It’s okay to use `as *mut _` but maybe type annotation (`let _: Box<…> =`) for sanity is best?

RRRadicalEdward: There does not have to be a `SecBuffer` there. The doc says that this function clears any buffer created by the security context. We do not know what type is behind that pointer. If we take FreeRDP for example, then `SecPkgInfoW /A` is deleted through `FreeContextBuffer`.

tools/sspi-ffi-attacker/src/attack.rs
```diff
@@ -0,0 +1,341 @@
...
+    if status != 0 {
+        panic!("QuerySecurityPackageInfoA failed: {}", status);
+    }
+
+    println!("{:?}", *pkg_info);
+    println!("{:?}", CStr::from_ptr((*pkg_info).Name));
+    println!("{:?}", CStr::from_ptr((*pkg_info).Comment));
+
+    let mut pc_packages = 0;
+    let mut packages: PSecPkgInfoA = null::<PSecPkgInfoA>() as *mut _;
```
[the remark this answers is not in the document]
RRRadicalEdward: Fixed it in the latest commit

RRRadicalEdward:
Hi @CBenoit! I've addressed your comments. Is there something wrong?

awakecoding (approved):

CBenoit (commented):
Sorry for the delay, I was sick

ffi/src/common.rs
RRRadicalEdward: It frees the memory the pointer points to, but doesn't call `drop()`. We can't identify what type `pv_context_buffer` contains as  `FreeContextBuffer` can be called with any buffer. 
I've checked the memory freeing with changes made in this PR with `cargo-valgrind` and there are no leaks.

CBenoit: Fair enough

PR #93, opened 2022-12-19
Allow hostname passing

RRRadicalEdward:
I've improved `sspi-rs` API and now we can pass the hostname directly into the `SspiContext`.

I added a new trait method `SspiEx::custom_set_hostname` and implemented it for all the security packages. I decided to not change the main `Sspi` trait because the original `SSPI` doesn't have such a method. On non-wasm targets, we still use `whoami` crate because it's needed for all other applications (like `FreeRDP`) that work with `sspi-rs` using FFI.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,6 +55,9 @@ portpicker = { version = "0.1.1", optional = true }
 num-bigint-dig = "0.8.1"
 tracing = { version = "0.1.37" }
 
+[target.'cfg(not(target_arch = "wasm32-unknown-unknown"))'.dependencies]
+whoami = "0.5"
+
 [target.'cfg(windows)'.dependencies]
 winreg = "0.10"
 winapi = { version = "0.3", features = ["sspi", "rpcdce", "impl-default", "timezoneapi", "wincrypt"] }
--- a/src/sspi/negotiate.rs
+++ b/src/sspi/negotiate.rs
@@ -222,6 +222,14 @@ impl SspiEx for Negotiate {
             NegotiatedProtocol::Ntlm(ntlm) => ntlm.custom_set_auth_identity(identity),
         }
     }
+
+    fn custom_set_hostname(&mut self, hostname: String) {
+        match &mut self.protocol {
+            NegotiatedProtocol::Pku2u(pku2u) => pku2u.custom_set_hostname(hostname),
+            NegotiatedProtocol::Kerberos(kerberos) => kerberos.custom_set_hostname(hostname),
+            NegotiatedProtocol::Ntlm(ntlm) => ntlm.custom_set_hostname(hostname),
+        }
+    }
 }
 
 impl Sspi for Negotiate {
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -55,6 +55,9 @@ portpicker = { version = "0.1.1", optional = true }
 num-bigint-dig = "0.8.1"
 tracing = { version = "0.1.37" }
 
+[target.'cfg(not(target_arch = "wasm32-unknown-unknown"))'.dependencies]
+whoami = "0.5"
+
```
CBenoit, line 60: Ideally, this dependency should be in `sspi-ffi` only and `custom_set_hostname` should be called with the result.

CBenoit (commented):
I was thinking the hostname would be a parameter when building  `CredSspClient` and the `whoami` dependency part of the `sspi-ffi` crate. Do you think that would be acceptable?

Cargo.toml
RRRadicalEdward: Done

RRRadicalEdward:
Hi @CBenoit! I added the option.

The hostname is only required for KDC/KerberosProxy. Also, I noticed that you wouldn't be able to negotiate a protocol in WASM. As KDC/KerberosProxy was behind the `network_client` feature that is not accessible on WASM(hardcoded usage of the `reqwest` client) in the `negotiate_protocol` function. So, I refactored the `negotiate_protocol` function to make it work on WASM. I created a separate `RequestClientFactory` trait, so you can pass your client in WASM. In FFI, it still uses the `reqwest` client.

tools/wasm-testcompile/src/lib.rs
```diff
@@ -1,14 +1,33 @@
 use sspi::internal::credssp;
+use sspi::negotiate::network_client::{NetworkClient, NetworkClientFactory};
 use sspi::ntlm::AuthIdentity;
 use wasm_bindgen::prelude::*;
 
+#[derive(Debug, Clone)]
+struct DumpNetworkClientFactory;
```
CBenoit, line 7: I’m unsure of the name. Do you mean "Dummy" here?

```suggestion
struct DummyNetworkClientFactory;
```

src/sspi/negotiate/network_client.rs
```diff
@@ -1,7 +1,14 @@
+use std::fmt::Debug;
```
CBenoit, line 1: I see you moved this module under `negotiate`. I understand this can be used at a few places (including kerberos) and it’s generic enough so I feel it could be at top level, what do you think?

src/sspi/negotiate.rs
```diff
@@ -43,29 +41,35 @@ pub trait ProtocolConfig: Debug {
...
+        protocol_config: Box<dyn ProtocolConfig + Send>,
+        package_list: Option<String>,
+        hostname: String,
+        get_network_client: Box<dyn NetworkClientFactory>,
+    ) -> Self {
         Self {
             protocol_config,
             package_list,
+            hostname,
+            network_client_generator: get_network_client,
```
CBenoit, line 59: I feel it would improve clarity to keep the same terminology at both places.
```suggestion
            network_client_factory: get_network_client,
```

CBenoit (commented):
Nice improvements, thank you!

tools/wasm-testcompile/src/lib.rs
RRRadicalEdward: Yeah 😄

RRRadicalEdward: Fixed

CBenoit (approved):
LGTM!

PR #94, opened 2022-12-20
Prepare next release

CBenoit:
FYI @RRRadicalEdward. I mostly cleaned up file organization. If you spot something weird, I would be happy to get your feedback. :slightly_smiling_face: 
Breaking changes due to the new organization should be minimal since most of the API was re-exported at the top level.
Once this PR is merged, I’ll proceed to publish the crate on `crates.io`.


RRRadicalEdward (approved):
LGTM

PR #95, opened 2023-01-20
 TSSSP module

RRRadicalEdward:
This PR adds support for using `sspi-rs` as a _TSSSP_ module. It adds the possibility to replace the native _credssp.dll_ with `sspi-rs` for the _mstsc_. _credssp.dll_ relies on the _Schannel_ and the _TSSSP_ and does not contain any protocol implementations. `sspi-rs`  uses Rust stack instead of _Schannel_ (`rustls` + `picky-rs`). 
I didn't create another _FFI_ for, but just extended the existing one and improved `SspiContext` by the new item.

Here are some important points about this implementation.
I've changed the _SSPI_ tables. Now in the reserved fields `EncryptMessage`/`DecryptMessage` functions are specified. This is not documented anywhere and was discovered during debugging. The _mstsc_ always uses the `reserved3` and `reserved4` fields of the functions table as `EncryptMessage` and `DecryptMessage` functions.
I also added support for the new attributes for the `QueryContextAttributesA/W` function:
- _SECPKG_ATTR_STREAM_SIZES_
- _SECPKG_ATTR_REMOTE_CERT_CONTEXT_
- _SECPKG_ATTR_NEGOTIATION_PACKAGE_
- _SECPKG_ATTR_PACKAGE_INFO_
- _SECPKG_ATTR_SERVER_AUTH_FLAGS_
- _SECPKG_ATTR_CERT_TRUST_STATUS_
- _SECPKG_ATTR_CONNECTION_INFO_

Small _FFI_ changes:
- pf_qop parameter of the `DecryptFunction` can be null if this library is used as a _CredSsp_ security package.
- Improved the `copy_to_c_sec_buffer` function: now it also copies the buffer type and handles when the buffer pointer is null.

At this point, we use the `rustls` for _TLS_ connection establishing and traffic encryption/decryption. I've placed the _TLS_ part into a separate module so we can easily add native _Schannel_ support.

Used Docs & References:
- [CredSsp](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cssp/85f57821-40bb-46aa-bfcb-ba9590b8fc30)
- [SChannel](https://learn.microsoft.com/en-us/windows-server/security/tls/tls-ssl-schannel-ssp-overview)
- [QueryContextAttributes (Schannel)](https://learn.microsoft.com/en-us/windows/win32/secauthn/querycontextattributes--schannel)
- [QueryContextAttributes (CredSSP)](https://learn.microsoft.com/en-us/windows/win32/secauthn/querycontextattributes--credssp)


src/credssp/sspi_cred_ssp/tls_connection.rs
```diff
@@ -0,0 +1,220 @@
+use std::io::{Read, Write};
```
CBenoit, line 1: For my information, do you think this module can compile for WASM? I don’t think rustls is fully WASM-compatible. I’m not sure [wasm-testcompile](https://github.com/Devolutions/sspi-rs/tree/master/tools/wasm-testcompile) is triggering code path in this module. Do you think you could extend it? (Unless it doesn’t make sense for us to support TSSSP in IronRDP of course.)

CBenoit (commented):
Thank you! Should we also bump the version?

ffi/src/sec_winnt_auth_identity.rs
```diff
@@ -80,6 +83,237 @@ pub struct SecWinntAuthIdentityEx2 {
...
+
+    let domain_len_ptr = user_len_ptr.add(8);
+    let domain_buffer_len = *domain_len_ptr as u32;
+
+    let creds_len_ptr = domain_len_ptr.add(8);
+    let creds_buffer_len = *creds_len_ptr as u32;
+
+    // header size + buffers size
+    64 /* size of SEC_WINNT_AUTH_IDENTITY_EX2 */ + user_buffer_len + domain_buffer_len + creds_buffer_len
+}
```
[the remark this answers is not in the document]
awakecoding: it's a fairly bad data structure with different memory layouts, it's very poorly documented: https://github.com/FreeRDP/FreeRDP/blob/master/winpr/libwinpr/sspi/sspi_winpr.c#L473

ffi/src/security_tables.rs
```diff
@@ -55,8 +55,8 @@ pub struct SecurityFunctionTableA {
     pub VerifySignature: VerifySignatureFn,
     pub FreeContextBuffer: FreeContextBufferFn,
     pub QuerySecurityPackageInfoA: QuerySecurityPackageInfoFnA,
-    pub Reserved3: *const c_void,
-    pub Reserved4: *const c_void,
+    pub Reserved3: EncryptMessageFn,
+    pub Reserved4: DecryptMessageFn,
```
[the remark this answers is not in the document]
RRRadicalEdward: Good idea. I added the comment about it in the code

ffi/src/sec_winnt_auth_identity.rs
```diff
@@ -80,6 +83,237 @@ pub struct SecWinntAuthIdentityEx2 {
...
+#[derive(Debug)]
+#[repr(C)]
+pub struct CredSspCred {
+    pub submit_type: CredSspSubmitType,
+    pub p_schannel_cred: *const c_void,
+    pub p_spnego_cred: *const c_void,
+}
+
+#[cfg(not(target_os = "windows"))]
+pub unsafe fn unpack_sec_winnt_auth_identity_ex2_a(_p_auth_data: *const c_void) -> Result<AuthIdentityBuffers> {
```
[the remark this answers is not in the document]
RRRadicalEdward: Yes, sure.  I removed it in the latest commit.

ffi/src/sec_winnt_auth_identity.rs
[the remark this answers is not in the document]
RRRadicalEdward: I added comments in the function. I should be more clear now

src/credssp/sspi_cred_ssp/tls_connection.rs
RRRadicalEdward: It doesn't make sense to support TSSSP in `IronRDP`. TSSSP is only expected to be used on _Windows_ when `sspi-rs` is used with `mstsc`.

awakecoding: @CBenoit as long as we can build this for Windows only without breaking wasm-builds with TSSSP disabled and other platforms, we're good. it's only going to be used for mstsc.exe injection for now

RRRadicalEdward: Let me make allow compiling it only on `Windows`

CBenoit: Great, thank you!

Cargo.toml
```diff
@@ -55,6 +55,8 @@ trust-dns-resolver = { version = "0.21.2", optional = true }
 portpicker = { version = "0.1.1", optional = true }
 num-bigint-dig = "0.8.1"
 tracing = { version = "0.1.37" }
+
+[target.'cfg(not(wasm))'.dependencies]
 rustls = { version = "0.20.7", features = ["dangerous_configuration"] }
```
CBenoit, line 60: I think you can target windows directly here

ffi/src/sec_winnt_auth_identity.rs
```diff
@@ -133,17 +133,20 @@ pub unsafe fn unpack_sec_winnt_auth_identity_ex2_a(_p_auth_data: *const c_void)
...
     let domain_len_ptr = user_len_ptr.add(8);
     let domain_buffer_len = *domain_len_ptr as u32;
 
+    // packet credentials length is placed after 16 bytes from the domain length
     let creds_len_ptr = domain_len_ptr.add(8);
     let creds_buffer_len = *creds_len_ptr as u32;
 
     // header size + buffers size
-    64 /* size of SEC_WINNT_AUTH_IDENTITY_EX2 */ + user_buffer_len + domain_buffer_len + creds_buffer_len
+    64 /* size of the SEC_WINNT_AUTH_IDENTITY_EX2 */ + user_buffer_len + domain_buffer_len + creds_buffer_len
```
CBenoit, line 149: Is there any external link that could be used as a reference here? Is https://github.com/FreeRDP/FreeRDP/blob/master/winpr/libwinpr/sspi/sspi_winpr.c#L473 the only reference?

src/credssp/sspi_cred_ssp/mod.rs
```diff
@@ -1,3 +1,4 @@
+#[cfg(not(target = "wasm32-unknown-unknown"))]
```
CBenoit, line 1: Likewise, if it’s only for windows maybe `#[cfg(windows)]` is better

RRRadicalEdward: Yeah. I'm working on it

Cargo.toml
RRRadicalEdward: Yes. It was my first try to fix it. I came up with a better idea of how to properly fix it and I'm working on it now

RRRadicalEdward: I added a separate feature `tsssp` in `sspi-rs` and `sspi-rs/ffi` that can only be enabled on Windows. The _TSSSP_ implementation is hidden behind the feature so sspi-rs can be built for the WASM target.  

Currently, CI builds sspi-rs with  `--all-features` flag on all systems which makes the build fail cause `tsssp` can be only enabled on Windows.  We should modify CI to build with the feature enabled only on Windows. I want to do it in a separate PR to not mix this PR, but I'm not sure if it's possible to merge a PR when there are some failed runners.

CBenoit:
I’m unsure whether using a new feature flag is the right choice. What about using `#[cfg(windows)]` directly?

RRRadicalEdward:
> I’m unsure whether using a new feature flag is the right choice (this requires enabling / disabling the feature depending on the targeted platform). What about using #[cfg(windows)] directly?

I think a new feature flag makes better intention. It makes code harder to read when there are a bunch of `#[cfg(target)]`. You would think  - are different features related if they compile under the same target?
 TSSSP use case is only to build `sspi-rs` and use it as `credssp.dll` in  `mstsc`. So I wouldn't be used in crates. 
I will change to use `#[cfg(windows)]` if you still insist on it.

CBenoit:
> TSSSP use case is only to build `sspi-rs` and use it as `credssp.dll` in `mstsc`. So I wouldn't be used in crates. I will change to use `#[cfg(windows)]` if you still insist on it.

Okay, leave it that way then.

> Currently, CI builds sspi-rs with --all-features flag on all systems which makes the build fail cause tsssp can be only enabled on Windows. We should modify CI to build with the feature enabled only on Windows. I want to do it in a separate PR to not mix this PR, but I'm not sure if it's possible to merge a PR when there are some failed runners.

As you said, this have to be fixed in this PR. I push to your branch and fix that if you want.
I can do that once remaining conversations are resolved.

ffi/Cargo.toml
```diff
@@ -14,17 +14,18 @@ crate-type = ["cdylib"]
...
 cfg-if = "0.1"
-sspi = { path = "..", features = ["network_client", "dns_resolver"] }
 libc = "0.2"
 num-traits = "0.2"
 whoami = "1.2.3"
 
 # For debugging only
 tracing = { version = "0.1.37", optional = true }
 tracing-subscriber = { version = "0.3.16", features = ["std", "fmt", "local-time", "env-filter"], optional = true }
+sspi = { path = "..", features = ["network_client", "dns_resolver"] }
```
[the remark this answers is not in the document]
RRRadicalEdward: Fixed

ffi/src/sec_winnt_auth_identity.rs
RRRadicalEdward: There is https://learn.microsoft.com/en-us/windows/win32/api/sspi/ns-sspi-sec_winnt_auth_identity_ex2 documentation. I added it as a reference in the code

CBenoit (approved):
Thank you! @RRRadicalEdward 
I’ll patch the ci file and merge

CBenoit:
Hi @RRRadicalEdward 

Apparently there are tests failing under Linux for sspi-ffi:

```
failures:

---- sec_pkg_info::tests::enumerate_security_packages_a stdout ----
thread 'sec_pkg_info::tests::enumerate_security_packages_a' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `5`', ffi/src/sec_pkg_info.rs:339:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- sec_pkg_info::tests::enumerate_security_packages_w stdout ----
thread 'sec_pkg_info::tests::enumerate_security_packages_w' panicked at 'assertion failed: `(left == right)`
  left: `4`,
 right: `5`', ffi/src/sec_pkg_info.rs:357:13
```

RRRadicalEdward:
Hi @CBenoit! I fixed the units tests, but CI fails due to a CI itself related issue

CBenoit:
Thank you! I’m looking into it

CBenoit:
When I run the command manually on Windows, I indeed get the error locally as well. It says the specified executable is not a valid Win32 application. :thinking: 
You can reproduce with `cargo test --manifest-path ffi/Cargo.toml`. Strangely, this is only happening with windows, on Linux and mac there is no problem :shrug:

CBenoit:
I don’t understand why it’s failing only on Windows. Since it’s not happening in CI only, if you find anything let me know! For now I asked our Devops to relax a bit our requirements for this check and I’ll merge in order to avoid blocking this PR longer.

CBenoit:
Fixed CI in https://github.com/Devolutions/sspi-rs/pull/98

PR #97, opened 2023-01-31
PKU2U: Fix a private key of a certificate extraction

RRRadicalEdward:
The  `winapi` crate has string literals without the NULL terminator and we haven't handled this case. It was lucky to extract a private key sometimes when there were zeros after the sting in memory.


CBenoit (approved):
Thank you! I wonder if this is something that should be patched upstream as well?

RRRadicalEdward:
>  I wonder if this is something that should be patched upstream as well?
Yeah. Do you want to do a minor version bump? I'd create a PR that bumps version

CBenoit:
> Yeah. Do you want to do a minor version bump? I'd create a PR that bumps the version

Maybe we can wait for the remaining known issues to be fixed? I would be okay with that :slightly_smiling_face:

PR #101, opened 2023-02-09
Kerberos Unconstrained Delegation Support

### LINKED ISSUE #81: Kerberos Unconstrained Delegation Support
I was wondering whether this library supports Kerberos Unconstrained delegation.

With GSSAPI unconstrained delegate works by requesting a forwardable TGT from the KDC (`kinit -f username@REALM.COM`). This forwardable ticket can then be delegated to the host by passing in the `GSS_C_DELEG_FLAG` flag with `gss_init_sec_context`. There's also the `GSS_C_DELEG_POLICY_FLAG` which is like `GSS_C_DELEG_FLAG` but it will instead only delegate if the `ok-as-delegate` flag is set on the account in AD.

On SSPI it will attempt to get a forwardable ticket when `InitializeSecurityContext` has been used with the `ISC_REQ_DELEGATE` flag. This is like the `GSS_C_DELEG_FLAG` in that it will only get a delegated ticket if the `ok-as-delegate` flag is set in AD. AFAIK there's no way in SSPI to get a delegate ticket if the account isn't trusted for unconstrained delegation.

What this looks like in practice is that when you authenticate to a host without a delegatable ticket you see the following from `klist`

```
Current LogonId is 0:0xb2594

Cached Tickets: (1)

#0>     Client: vagrant-domain @ WSMAN.ENV
        Server: http/test.wsman.env @ WSMAN.ENV
        KerbTicket Encryption Type: AES-256-CTS-HMAC-SHA1-96
        Ticket Flags 0xa10000 -> renewable pre_authent name_canonicalize
        Start Time: 11/19/2022 9:08:58 (local)
        End Time:   11/19/2022 19:08:07 (local)
        Renew Time: 0
        Session Key Type: AES-256-CTS-HMAC-SHA1-96
        Cache Flags: 0x8 -> ASC
        Kdc Called:
```

When you authenticate with a delegatable ticket you see the following

```
Current LogonId is 0:0xb5caf

Cached Tickets: (1)

#0>     Client: vagrant-domain @ WSMAN.ENV
        Server: http/test.wsman.env @ WSMAN.ENV
        KerbTicket Encryption Type: AES-256-CTS-HMAC-SHA1-96
        Ticket Flags 0x60a10000 -> forwardable forwarded renewable pre_authent name_canonicalize
        Start Time: 11/19/2022 9:09:40 (local)
        End Time:   11/19/2022 19:09:35 (local)
        Renew Time: 0
        Session Key Type: AES-256-CTS-HMAC-SHA1-96
        Cache Flags: 0x8 -> ASC
        Kdc Called:
```

Notice that the `forwarded` ticket flag is set for a delegated scenario which allows Windows to reuse that Kerberos ticket for any further network hops. Say connecting to an SMB share can now re-use those credentials.

So ultimately my questions are:

+ Does this library support the `ISC_REQ_DELEGATE` flag in that it will get a service ticket the remote host can forward as needed
+ If so, does it do it unilaterally or does it honour the `ok-as-delegate` setting set on the account in question

RRRadicalEdward:
This PR adds support for the `ISC_REQ_DELEGATE` flag. See https://github.com/Devolutions/sspi-rs/issues/81 for more information. 
Also, I replace some numbers with actual flags (`KdcOptions`, `ApOptions`) and removed usage of hard coded the `SERVICE_NAME` constant(It's parsed from input now).

Closes #81


CBenoit (approved):
LGTM!

PR #103, opened 2023-02-20
fix(sspi-rs): Improve mapping Kerberos errors into SSPI errors

### LINKED ISSUE #83: Better Error for Clock Skew
When Kerberos fails due to the clock being out of sync the AP-REQ errors with `KRB_AP_ERR_SKEW` but sspi-rs returns `SEC_E_INTERNAL_ERROR 0x80090304` during the `InitializeSecurityContext` call. It would be nice if it returned a more specific error like `SEC_E_TIME_SKEW 0x80090324` instead.

RRRadicalEdward:
This PR fixes #83

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,10 +47,10 @@ serde_derive = "1.0"
 url = "2.2.2"
 reqwest = { version = "0.11", features = ["blocking", "rustls-tls", "rustls-tls-native-roots"], optional = true, default-features = false }
 picky = { version = "7.0.0-rc.3", default-features = false }
-picky-krb = { version = "0.5.0" }
-picky-asn1 = { version = "0.7.0", features = ["chrono_conversion"] }
-picky-asn1-der = { version = "0.4.0" }
-picky-asn1-x509 = { version = "0.9.0", features = ["pkcs7"] }
+picky-krb = { git = "https://github.com/Devolutions/picky-rs.git", rev = "871e0f7ab01cb98e732c94b78e8241ac72e219cb" }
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -47,10 +47,10 @@ serde_derive = "1.0"
...
 reqwest = { version = "0.11", features = ["blocking", "rustls-tls", "rustls-tls-native-roots"], optional = true, default-features = false }
 picky = { version = "7.0.0-rc.3", default-features = false }
-picky-krb = { version = "0.5.0" }
-picky-asn1 = { version = "0.7.0", features = ["chrono_conversion"] }
-picky-asn1-der = { version = "0.4.0" }
-picky-asn1-x509 = { version = "0.9.0", features = ["pkcs7"] }
+picky-krb = { git = "https://github.com/Devolutions/picky-rs.git", rev = "871e0f7ab01cb98e732c94b78e8241ac72e219cb" }
+picky-asn1 = { git = "https://github.com/Devolutions/picky-rs.git", rev = "871e0f7ab01cb98e732c94b78e8241ac72e219cb", features = ["chrono_conversion"] }
+picky-asn1-der = { git = "https://github.com/Devolutions/picky-rs.git", rev = "871e0f7ab01cb98e732c94b78e8241ac72e219cb" }
+picky-asn1-x509 = { git = "https://github.com/Devolutions/picky-rs.git", rev = "871e0f7ab01cb98e732c94b78e8241ac72e219cb", features = ["pkcs7"] }
```
CBenoit, line 53: Do we need to bump other crates?

CBenoit (commented):
Looks good to me, but we should remove git references in dependencies. Anything I can do for you?

Cargo.toml
RRRadicalEdward: It's already been published. I pushed a commit that fixes it

CBenoit (approved):
Thank you

PR #105, opened 2023-02-22
feat: allow to enable logging in runtime.

### LINKED ISSUE #85: Support for trace logs at runtime
There is limited trace logging available with https://github.com/Devolutions/sspi-rs/pull/73 but it requires the user to have a binary compiled with `debug_mode` enabled. It would be great if there was a feature like [KRB5_TRACE](https://web.mit.edu/kerberos/krb5-latest/doc/admin/troubleshoot.html) that MIT krb5 supports. This is extremely useful when dealing with Kerberos problems and trying to figure out why things are not working. The benefit is that it's enabled even in the release builds so you don't need to tell users to replace this dll with something else, they just need to set an env var.

I have no idea if Rust has some general logging mechanism that could be used instead but ultimately what I am hoping for is a way to debug broken scenarios and getting logs from a user is pretty critical for this.

RRRadicalEdward:
The logging level is controlled by the `SSPI_LOG_LEVEL` environment variable. The output log file is specified by `SSPI_LOG_PATH`. 
Also, I added more logs similar to what https://github.com/krb5/krb5 logs.

This PR fixes #85.


RRRadicalEdward:
CI fails on some targets cause I removed the _debug_mode_  compilation feature in _ffi_

src/pku2u/cert_utils/validation.rs
```diff
@@ -2,32 +2,44 @@ use num_bigint_dig::BigUint;
...
 pub fn validate_server_p2p_certificate(signed_data: &SignedData) -> Result<RsaPublicKey> {
     let certificates = &signed_data.certificates.0 .0;
 
     if let Some(certificate) = certificates.iter().next() {
         let cert: Certificate = match certificate {
             CertificateChoices::Certificate(cert) => picky_asn1_der::from_bytes(&cert.0)?,
-            _ => {
+            _cert => {
+                #[cfg(feature = "logging")]
+                error!("Server sent unsupported certificate format: {:?}", _cert);
```
CBenoit, line 21: The tracing way:
```suggestion
                error!(cert = ?_cert, "Server sent unsupported certificate format");
```

src/pku2u/cert_utils/validation.rs
```diff
@@ -2,32 +2,44 @@ use num_bigint_dig::BigUint;
...
 
         let public_key = match cert.tbs_certificate.subject_public_key_info.subject_public_key {
             PublicKey::Rsa(rsa) => rsa,
-            _ => {
+            _public_key => {
+                #[cfg(feature = "logging")]
+                error!(
+                    "Server sent unsupported public key type: {:?}. Only RSA keys supported",
+                    _public_key
+                );
```
CBenoit, line 37: The tracing way:
```suggestion
                error!(
                    public_key: ?_public_key,
                    "Server sent unsupported public key type: {:?}. Only RSA keys supported",
                );
```

Note that you should remove the `_` prefix if you remove the `logging` feature gate.

src/pku2u/cert_utils/win_extraction.rs
```diff
@@ -58,6 +60,12 @@ fn decode_private_key(mut buffer: impl Read) -> Result<PrivateKey> {
     let rsa_key_blob = BcryptRsaKeyBlob::from_read(&mut buffer)?;
 
     if rsa_key_blob.magic != BCRYPT_RSAFULLPRIVATE_MAGIC {
+        #[cfg(feature = "logging")]
+        debug!(
+            "Invalid RSA key blob magic. Expected {} but got {}",
+            BCRYPT_RSAFULLPRIVATE_MAGIC, rsa_key_blob.magic
+        );
```
CBenoit, line 67: The tracing way:
```suggestion
        debug!(
            expected = BCRYPT_RSAFULLPRIVATE_MAGIC,
            actual = rsa_key_blob.magic,
            "Invalid RSA key blob magic",
        );
```

src/pku2u/cert_utils/win_extraction.rs
```diff
@@ -157,6 +168,12 @@ unsafe fn export_certificate_private_key(cert: *const CERT_CONTEXT) -> Result<Pr
     );
 
     if status == 0 || private_key_handle == 0 {
+        #[cfg(feature = "logging")]
+        error!(
+            "Cannot acquire certificate private key handle. Status: {:x?}. Private key handle: {}",
+            status, private_key_handle
+        );
```
CBenoit, line 175: The tracing way:
```suggestion
        #[cfg(feature = "logging")]
        error!(
            ?status, private_key_handle,
            "Cannot acquire certificate private key handle",
        );
```

src/pku2u/cert_utils/win_extraction.rs
```diff
@@ -240,6 +260,12 @@ unsafe fn extract_client_p2p_certificate(cert_store: *mut c_void) -> Result<(Cer
             continue;
         }
 
+        #[cfg(feature = "logging")]
+        debug!(
+            "Found suitable client p2p certificate. Serial number: {:?}",
+            cert.tbs_certificate.serial_number.0
+        );
```
CBenoit, line 267: The tracing way:
```suggestion
        #[cfg(feature = "logging")]
        debug!(
            serial_number = ?cert.tbs_certificate.serial_number.0,
            "Found suitable client p2p certificate",
        );
```

src/pku2u/extractors.rs
```diff
@@ -87,7 +92,15 @@ pub fn extract_server_dh_public_key(signed_data: &SignedData) -> Result<Vec<u8>>
...
-        _ => return Err(Error::new(ErrorKind::InvalidToken, "unexpected content info".into())),
+        _content_value => {
+            #[cfg(feature = "logging")]
+            error!(
+                "The server has sent KDC DH key info in unsupported format: {:?}. Only ContentValue::OctetString is supported",
+                _content_value
+            );
+
+            return Err(Error::new(ErrorKind::InvalidToken, "unexpected content info".into()));
+        }
```
CBenoit, line 103: Similar to above

src/pku2u/generators.rs
```diff
@@ -158,6 +163,13 @@ pub fn generate_signer_info(p2p_cert: &Certificate, digest: Vec<u8>, private_key
             )
         })?;
 
+    #[cfg(feature = "logging")]
+    trace!(
+        "Pku2u signed attributes: plain {:?}, signed (signature): {:?}",
+        encoded_signed_attributes,
+        signature
+    );
```
CBenoit, line 171: Similar to above

src/pku2u/mod.rs
```diff
@@ -198,16 +214,21 @@ impl Pku2u {
...
 
+    #[cfg_attr(feature = "logging", instrument(level = "debug", ret, fields(state = self.state.as_ref()), skip(self, _flags)))]
     fn encrypt_message(
         &mut self,
         _flags: EncryptionFlags,
         message: &mut [SecurityBuffer],
         sequence_number: u32,
     ) -> Result<SecurityStatus> {
+        #[cfg(feature = "logging")]
+        trace!("{:?}", self.encryption_params);
```
CBenoit, line 230: Tracing way would be:
```suggestion
        #[cfg(feature = "logging")]
        trace!(encryption_params = ?self.encryption_params);
```
(you don’t need a (format) message string at all)

src/pku2u/mod.rs
```diff
@@ -271,7 +282,11 @@ impl Sspi for Pku2u {
         Ok(SecurityStatus::Ok)
     }
 
+    #[cfg_attr(feature = "logging", instrument(level = "debug", ret, fields(state = self.state.as_ref()), skip(self, _sequence_number)))]
     fn decrypt_message(&mut self, message: &mut [SecurityBuffer], _sequence_number: u32) -> Result<DecryptionFlags> {
+        #[cfg(feature = "logging")]
+        trace!("{:?}", self.encryption_params);
```
CBenoit, line 288: Similar to above

src/pku2u/mod.rs
```diff
@@ -290,19 +305,12 @@ impl Sspi for Pku2u {
...
-            return Err(Error::new(
-                ErrorKind::DecryptFailure,
-                "No encryption key provided".into(),
-            ));
-        };
+        let key = get_encryption_key(&self.encryption_params)?;
         let key_usage = self.encryption_params.sspi_decrypt_key_usage;
 
+        #[cfg(feature = "logging")]
+        debug!("Decryption key usage: {}", key_usage);
```
CBenoit, line 312: Similar to above

ffi/src/logging.rs
```diff
@@ -0,0 +1,76 @@
...
+use std::path::PathBuf;
+use std::sync::Once;
+
+use tracing::error;
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::EnvFilter;
+
+static SETUP: Once = Once::new();
+
+const SSPI_DBG_LOG_PATH_ENV: &str = "SSPI_LOG_FILE_PATH";
```
CBenoit, line 11: Could you name it `SSPI_LOG_PATH`? Also rename the const variable name.
```suggestion
const SSPI_LOG_PATH_ENV: &str = "SSPI_LOG_PATH";
```

ffi/src/logging.rs
```diff
@@ -0,0 +1,76 @@
...
+
+pub fn setup_logger() {
+    SETUP.call_once(|| {
+        let path = if let Ok(path) = std::env::var(SSPI_DBG_LOG_PATH_ENV) {
+            PathBuf::from(path)
+        } else if let Ok(path) = std::env::var(HOME_PATH_ENV) {
+            let mut path = PathBuf::from(path);
+            path.push("sspi-rs.log");
+            path
+        } else {
```
CBenoit, line 27: In this case, we should probably not register any log subscriber if the `SSPI_LOG_PATH` env variable is not set. Or, do we want logs unconditionally? If we do, we probably don’t want to use user’s home directory by default anymore. This would be more appropriate I think: https://docs.rs/dirs/latest/dirs/fn.data_dir.html
But, as for me, I believe logs should be opt-in.

Input? @awakecoding @thenextman

CBenoit (commented):
Thank you @RRRadicalEdward! I believe this will really help with future debugging.

Out of curiosity, could you send us a log file sample so we can see how it looks like? :smile: 

@awakecoding @thenextman input? Any security concern regarding logging? I don’t know SSPI details that much.

PS: I’ll fix CI once all comments are addressed! :slightly_smiling_face:

ffi/src/logging.rs
thenextman: Yes, logs should definitely be opt-in

thenextman:
> Thank you @RRRadicalEdward! I believe this will really help with future debugging.
> 
> Out of curiosity, could you send us a sample log file so we can see how it looks like? 😄
> 
> @awakecoding @thenextman input? Any security concern regarding logging? I don’t know SSPI details that much.
> 
> PS: I’ll fix CI once all comments are addressed! 🙂

There are definite security concerns with trace logging; all logging should be opt-in and just enabled for debugging.

Care must be taken when handling log files; for example in FreeRDP we allow capturing session trace logs. Since we enable debug logging for NTLM in this case; we must be careful when instructing users and harvesting log files (i.e. they should only be enabled when needed, not be posted on public forum, only sent by secure message)

I don't see us exposing this option in any of our products initially, I see it more as a development tool.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ exclude = [
 default = []
 network_client = ["dep:reqwest", "dep:portpicker"]
 dns_resolver = ["dep:trust-dns-resolver"]
+logging = ["tracing"]
```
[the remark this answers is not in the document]
RRRadicalEdward: Is ok that tracing would be compiled for users who don't even want to use logs?
It's why I placed it behind the feature

CBenoit:
> 

Do you think we should keep all log behind the compile-time feature flag initially (the way it is on master)?
That way, there is no risk of a malicious actor setting the environment variable to extract logs.

thenextman:
> Do you think we should keep all log behind the compile-time feature flag initially (the way it is on master)?
> That way, there is no risk of a malicious actor setting the environment variable to extract logs as long as we don’t provide a debug-enabled build.

IMO it should be available generally, like KRB_TRACE. However I'm fine to leave it for debug builds, at least for now :)

Cargo.toml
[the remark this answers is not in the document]
CBenoit: It’s generally ok, most crates with logging do that:
- https://github.com/tokio-rs/mio/blob/7ed74bf478230a0cfa7543901f6be6df8bb3602e/Cargo.toml#L46
- https://github.com/hyperium/hyper/blob/3fb59919941d3145be6d84dab85d222ea0e7664b/Cargo.toml#L33

Note that they only import the macro-providing library (either `log` or `tracing`).

It’s very lightweight and costs practically nothing. However, log events are still emitted and may be intercepted. That’s why we should be careful with what is logged, and decide whether we should totally disable everything at compile-time (even in the ffi crate, that is the way it is an master currently).

CBenoit:
> > Do you think we should keep all log behind the compile-time feature flag initially (the way it is on master)?
> > That way, there is no risk of a malicious actor setting the environment variable to extract logs as long as we don’t provide a debug-enabled build.
> 
> IMO it should be available generally, like KRB_TRACE. However I'm fine to leave it for debug builds, at least for now :)

Okay! @RRRadicalEdward already removed the feature-gating in the FFI crate, so it will be generally available :+1: 

You can go forward fixing the comments @RRRadicalEdward :slightly_smiling_face:

Cargo.toml
```diff
@@ -56,7 +57,7 @@ uuid = { version = "1.1", features = ["v4"] }
 trust-dns-resolver = { version = "0.21.2", optional = true }
 portpicker = { version = "0.1.1", optional = true }
 num-bigint-dig = "0.8.1"
-tracing = { version = "0.1.37" }
+tracing = { version = "0.1.37", features = ["attributes"], optional = true }
```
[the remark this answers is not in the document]
RRRadicalEdward: Yeah. I fixed it

ffi/src/sec_pkg_info.rs
```diff
@@ -7,6 +7,7 @@ use libc::{c_uint, c_ulong, c_ushort};
 use sspi::{enumerate_security_packages, PackageInfo, KERBEROS_VERSION};
 #[cfg(windows)]
 use symbol_rename_macro::rename_symbol;
+use tracing::instrument;
```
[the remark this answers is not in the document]
RRRadicalEdward: Fixed

src/credssp/mod.rs
```diff
@@ -920,6 +969,9 @@ impl CredSspContext {
         let expected_public_key = compute_sha256(&data);
 
         if expected_public_key.as_ref() != decrypted_public_key.as_slice() {
+            #[cfg(feature = "logging")]
+            error!("Expected and decrypted public key hash are not the same");
```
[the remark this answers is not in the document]
RRRadicalEdward: I removed the feature gate so it's no needed

ffi/src/sec_handle.rs
```diff
@@ -130,6 +132,8 @@ pub(crate) unsafe fn p_ctxt_handle_to_sspi_context(
         }
         let name = security_package_name.expect("security package name must be provided");
 
+        info!("Creating {:?} context", name);
```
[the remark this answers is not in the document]
RRRadicalEdward: Fixed

Cargo.toml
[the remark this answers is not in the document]
RRRadicalEdward: Ok. I removed the feature gate

src/pku2u/cert_utils/validation.rs
RRRadicalEdward: Fixed

src/pku2u/cert_utils/win_extraction.rs
RRRadicalEdward: Fixed

src/pku2u/extractors.rs
RRRadicalEdward: Fixed

src/pku2u/generators.rs
RRRadicalEdward: Fixed

src/pku2u/cert_utils/validation.rs
RRRadicalEdward: Fixed

src/pku2u/cert_utils/win_extraction.rs
RRRadicalEdward: Fixed

src/pku2u/mod.rs
RRRadicalEdward: Fixed

src/pku2u/mod.rs
RRRadicalEdward: Fixed

src/pku2u/mod.rs
RRRadicalEdward: Fixed

ffi/src/logging.rs
RRRadicalEdward: Fixed

ffi/src/logging.rs
RRRadicalEdward: I fixed it. Now `SSPI_LOG_PATH` has to be set for logging

RRRadicalEdward:
@CBenoit  I resolved the comments. Please have a look one more time 🙂

CBenoit (approved):
Thank you! Ready to merge :smile:

PR #106, opened 2023-02-28
refactor: logging improvements

RRRadicalEdward:
This PR completes [the previous PR](https://github.com/Devolutions/sspi-rs/pull/105) by enhancing the security of logging. 

This PR contains the next improvements: 
- Add `Secret<T>` with custom `Debug`/`Display` implementations and `Zeroize` support. It has a generic parameter because in _sspi-rs_ there are a few types for passwords: `String` and ` Vec<u8>`.
- Add a separate structure `SecretPrivateKey` because `PrivateKey` does not implement `Zeroize`.
- Move `AuthIdentity` and `AuthIdentityBuffers` from `ntlm`  into a separate module. Those two structures are used in every implemented protocol. So, I decided to create a separate space for them. 
- Custom `Debug` trait implementation for` SecurityBuffer`/`AuthIdentityBuffers`.


src/secret.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl<T: Zeroize> Secret<T> {
+    pub fn new(inner: T) -> Self {
+        Self(inner)
+    }
+}
+
+impl<T: Zeroize> fmt::Debug for Secret<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "secret")?;
```
CBenoit, line 18:
```suggestion
        write!(f, "Secret")?;
```

src/secret.rs
```diff
@@ -0,0 +1,85 @@
...
+
+impl SecretPrivateKey {
+    pub fn new(inner: PrivateKey) -> Self {
+        Self(inner)
+    }
+}
+
+impl fmt::Debug for SecretPrivateKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "secret private key")?;
```
CBenoit, line 61:
```suggestion
        write!(f, "SecretPrivateKey")?;
```

ffi/src/sec_handle.rs
```diff
@@ -1029,15 +1029,15 @@ pub unsafe extern "system" fn ChangeAccountPasswordW(
...
+        let mut new_password = Secret::new(c_w_str_to_string(psz_new_password));
 
         ChangeAccountPasswordA(
             security_package_name.as_ptr() as *mut _,
             domain.as_ptr() as *mut _,
             username.as_ptr() as *mut _,
-            password.as_ptr() as *mut _,
-            new_password.as_ptr() as *mut _,
+            password.as_mut().as_ptr() as *mut _,
+            new_password.as_mut().as_ptr() as *mut _,
```
CBenoit, line 1040: I believe `as_mut_ptr()` should be used here. Do you know if there is a reason for not doing so?

CBenoit (commented):
Could you also also update dependencies on picky crates please?

src/secret.rs
RRRadicalEdward: Fixed

src/secret.rs
RRRadicalEdward: Fixed

ffi/src/sec_handle.rs
RRRadicalEdward: I added using `as_mut_ptr()` here

CBenoit (approved):
Thank you! This looks good! :smile:

PR #107, opened 2023-02-28
feat: AuthIdentity Workstation API

### LINKED ISSUE #36: API to set Workstation for AuthIdentity
In some case, may need  Workstation for negotiate, is there any interface for it.

RRRadicalEdward:
_Workstation_ is _NetBIOS_ name of your workstation.

I've added the `NtlmConfig` structure with the _workstation_ field. Now, if the user sets the _workstation_, it'll be sent to the remote server in the `Negotiate` message during the authorization.
In the _FFI_ user can set the _workstation_ using the `SetCredentialsAttributesA/W` function with the `SECPKG_CRED_ATTR_NAMES` attribute. If the _workstation_ is not provided, then `whoami::hostname()` is used.

This PR fixes #36.


CBenoit (approved):
Thank you!

PR #111, opened 2023-03-21
fixes dns_query_srv_records() for MacOS

ibeckermayer:
Makes it possible to call the MacOS version of dns_query_srv_records() from an existing tokio runtime.

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -133,7 +133,6 @@ cfg_if::cfg_if! {
     if #[cfg(any(target_os="macos", target_os="ios"))] {
         use std::time::Duration;
         use std::thread;
-        use std::sync::mpsc::channel;
         use tokio::time::timeout;
         use tokio::runtime;
         use futures::stream::{StreamExt};
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -215,16 +215,11 @@ cfg_if::cfg_if! {
                 Ok(handle) => {
                     // Tokio runtime already exists, cannot block again on the same thread.
                     // Spawn a new thread to run the blocking code.
-                    let (sender, receiver) = channel();
                     let name = name.to_owned();
                     thread::spawn(move || {
-                        let dns_records = handle.block_on(query_with_timeout(name, query_timeout));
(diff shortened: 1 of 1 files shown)


thenextman (changes requested):
Thanks for your contribution! The build is broken for non-Apple platforms, I added a comment to that effect.

thenextman:
LGTM. Paging @CBenoit to take a look as `tokio` is out of my wheelhouse.

src/dns.rs
```diff
@@ -208,9 +207,35 @@ cfg_if::cfg_if! {
...
+                    let (sender, receiver) = channel();
+                    let name = name.to_owned();
+                    thread::spawn(move || {
+                        let dns_records = handle.block_on(query_with_timeout(name, query_timeout));
+                        // Send the dns_records back to the main thread.
+                        sender.send(dns_records).unwrap();
+                    });
+
+                    // Wait for the dns_records to be sent back from the blocking thread.
+                    receiver.recv().unwrap()
```
CBenoit, line 227: I would prefer we join the thread and retrieve the result that way instead of using a mpsc channel.
https://doc.rust-lang.org/std/thread/struct.JoinHandle.html#method.join
See also the last example here: https://doc.rust-lang.org/std/thread/fn.spawn.html
What do you think?

CBenoit (commented):
Looks pretty good to me, thank you! I have a minor comment, could you have a look at it?

CBenoit (approved):
Great! Many thanks 😄

thenextman (approved):
LGTM

CBenoit:
Hi @thenextman , could you remove the change requested state? It appears I can’t merge as long as one maintainer review is in this state

src/dns.rs
```diff
@@ -132,10 +132,12 @@ cfg_if::cfg_if! {
 cfg_if::cfg_if! {
     if #[cfg(any(target_os="macos", target_os="ios"))] {
         use std::time::Duration;
+        use std::thread;
+        use std::sync::mpsc::channel;
```
CBenoit, line 136:
```suggestion
```

PR #114, opened 2023-03-27
fix(ntlm): handling of authenticate message without random encrypted session key

### LINKED ISSUE #113: NTLM without random Session Key
Hi there,

I'm implementing "NTLM over HTTP" and use sspi-rs to process the NTLM Messages.
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-ntht/f09cf6e1-529e-403b-a8a5-7368ee096a6a

I'm using actix-web as web-server and chrome as NTLM client.

So far the process works until a valid looking NTLMSSP_AUTH Message (Step 3), but it does not have a "Random Session Key" inside.

I've traced the messages, and it seems, that Chrome sends the NEGOTIATE message without "Negotiate Key Exchange" Flag, so it is valid, that the Random Session Key is not inside the AUTHENTICATE message.

However, the "read_header" function of authenticate.rs stumbles over this line:

https://github.com/Devolutions/sspi-rs/blob/master/src/ntlm/messages/server/authenticate.rs#L96
```rs
if encrypted_random_session_key.buffer.len() != ENCRYPTED_RANDOM_SESSION_KEY_SIZE {
    return Err(crate::Error::new(
        crate::ErrorKind::InvalidToken,
        String::from("Invalid encrypted random session key"),
    ));
}
```

hence
```
pub const ENCRYPTED_RANDOM_SESSION_KEY_SIZE: usize = 16;
```

I'm wondering:
- should the session-key size or the check at all here depend on the NEGOTIATE_KEY_EXCHANGE flag?
https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/f9e6fbc4-a953-4f24-b229-ccdcc213b9ec
> If NTLMSSP_NEGOTIATE_KEY_EXCH is set, the server MUST check if client supplied a valid EncryptedRandomSessionKey in the AUTHENTICATE_MESSAGE (section 2.2.1.3); otherwise, the server MUST return SEC_E_INVALID_TOKEN.

- do I have to modify the AUTHENTICATE message before passing it to ntlm, like calculating the encrypted session key manually on serverside?

Any comments are highly appreciated :)

ndrsg:
make the encrypted random session key optional for NTLM

check it only if the negotiation-flag is set

Closes #113


RRRadicalEdward (commented):
Great work!
Please also run `cargo fmt` to format the code.
I also want to test the changes just in case. I'll return with the results 🙂

src/ntlm/messages/server/authenticate.rs
```diff
@@ -186,8 +186,15 @@ fn process_message_fields(
...
 
-    let mut encrypted_random_session_key = [0x00; ENCRYPTED_RANDOM_SESSION_KEY_SIZE];
-    encrypted_random_session_key.clone_from_slice(message_fields.encrypted_random_session_key.buffer.as_ref());
+    let mut encrypted_random_session_key: Option<[u8; ENCRYPTED_RANDOM_SESSION_KEY_SIZE]> = if message_fields.encrypted_random_session_key.buffer.len() == 0 {
+      None
+    } else {
+      // if the client sent a random session key, the length is expected to be ENCRYPTED_RANDOM_SESSION_KEY_SIZE = 16 bytes.
+      // otherwise clone_from_slice will fail.
+      let mut encrypted_random_session_key = [0x00; ENCRYPTED_RANDOM_SESSION_KEY_SIZE];
+      encrypted_random_session_key.clone_from_slice(message_fields.encrypted_random_session_key.buffer.as_ref());
```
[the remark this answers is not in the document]
ndrsg: ok sure! you want it squashed / force-pushed or a new commit?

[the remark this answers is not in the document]
RRRadicalEdward: Let it be a new commit

RRRadicalEdward:
@ndrsg I tested your changes, and it works properly🎉
The CI fails cause there are few clippy warnings. Please fix them, and then this PR will be ready to merge 🙂

RRRadicalEdward:
@CBenoit do you want to take a look?

RRRadicalEdward (approved):

src/ntlm/messages/server/authenticate.rs
```diff
@@ -186,8 +186,20 @@ fn process_message_fields(
...
+      0 => None,
+      ENCRYPTED_RANDOM_SESSION_KEY_SIZE => {
+        let mut encrypted_random_session_key = [0x00; ENCRYPTED_RANDOM_SESSION_KEY_SIZE];
+        encrypted_random_session_key.clone_from_slice(message_fields.encrypted_random_session_key.buffer.as_ref());
+        Some(encrypted_random_session_key)
+      },
+      key_len => {
+        return Err(crate::Error::new(
+          crate::ErrorKind::InvalidToken,
+          format!("Encrypted random session key has wrong length. Expected {ENCRYPTED_RANDOM_SESSION_KEY_SIZE} bytes, got {key_len} bytes."),
```
CBenoit, line 199: Nit: if I’m not mistaken, this `format!` is not required

```suggestion
          "Encrypted random session key has wrong length. Expected {ENCRYPTED_RANDOM_SESSION_KEY_SIZE} bytes, got {key_len} bytes.",
```

CBenoit (approved):
Thank you! This looks good to me! :smiley: 
I found one nit, I’ll leave you some time if you want to address it and merge afterward :slightly_smiling_face: 

Thank you too @RRRadicalEdward for the follow up!

src/ntlm/messages/server/authenticate.rs
ndrsg: Hi @CBenoit , thanks for the suggestion!

The ```format!``` is required to get a String (not string-slice) and also to insert the variables ```ENCRYPTED_RANDOM_SESSION_KEY_SIZE``` and ```key_len``` into the error-message.

Removing it, leads to ```expected String, found &str```.
Replacing it with ```String::from()``` leads to an error-message with the variable-names inserted instead of their values.

CBenoit: Oh, I’m sorry I missed the format parameter!

Regarding the `expected String, found &str` error, this is probably because you are based on a slightly older version of the codebase. This was changed recently to accept both: https://github.com/Devolutions/sspi-rs/pull/112/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R1776

Anyway, since you are formatting text, you don’t need to change anything, no need to rebase in this case :slightly_smiling_face:

PR #124, opened 2023-06-02
fix(ffi): Fix stack corruption on decrypt message

RRRadicalEdward:
On Linux `c_ulong` is 8 bytes but `FreeRDP` provides a pointer to the 4-byte number. So, if we use c_ulong, we overwrite 4 bytes on the stack during the assigning. This PR fixes it. 
I refactored code to use just `c_uint` which is always 4 bytes wide instead of  `c_ulong`.


awakecoding:
fixed by https://github.com/Devolutions/sspi-rs/pull/125

PR #126, opened 2023-06-05
eliminate c_ulong, c_uint usage

awakecoding:
yeet c_ulong, c_uint usage into the sun, including conditional Windows/non-Windows definitions meant to ensure an unsigned integer of 32-bits on all platforms, and use u32 consistently for all platforms. ULONG / unsigned long on Windows is always 32-bit and FreeRDP maintains this ABI property everywhere, there is no point in going with the unsigned long C type which sometimes grows to 64-bits like on Linux.


thenextman (approved):
LGTM

PR #129, opened 2023-06-08
Fix usage of Kerberos use-session-key flag with KRB5 U2U

awakecoding:
KRB5 U2U (1.2.840.113554.1.2.2.3) requires the use-session-key flag, but FreeRDP doesn't explicitly set ISC_REQ_USE_SESSION_KEY in the InitializeSecurityContext flags. mstsc always sets ISC_REQ_USE_SESSION_KEY, but also always selects KRB5 U2U. With FreeRDP + Windows SSPI, it looks like setting ISC_REQ_USE_SESSION_KEY forces KRB5 U2U, while not setting it defaults to regular Kerberos. The FreeRDP WinPR Kerberos implementation sets the use-session-key flag based on the negotiated mech id (U2U or not) rather than expect the ISC_REQ_USE_SESSION_KEY flag to be passed to the Negotiate module. We're going for a similar fix on our side, and I suspect we may only support KRB5 U2U at this point anyway.


thenextman (approved):
LGTM

PR #131, opened 2023-06-29
Kerberos fixes

RRRadicalEdward:
This PR fixes issues with _Kerberos_ that Richard has faced. 
- Remove using of `SecPkgCredentialsKdcProxySettingsA` as _Microsoft_ doesn't have this structure and uses `SecPkgCredentialsKdcProxySettingsW`even for _ANSI_ functions. This fixes setting _KdcProxy_ URL on non-_Windows_ machines. 
- Add setting of the hostname in _Kerberos_ config when it's initialized from environment variables.

### DIFF
--- a/ffi/src/credentials_attributes.rs
+++ b/ffi/src/credentials_attributes.rs
@@ -1,3 +1,8 @@
+use std::mem::size_of;
+use std::slice::from_raw_parts;
+
+use libc::c_void;
+
 use crate::sspi_data_types::{SecChar, SecWChar};
 
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


CBenoit (approved):
I guess it’s not relevant anymore, but it looks like both `SecPkgCredentialsKdcProxySettingsA` and `SecPkgCredentialsKdcProxySettingsW` were identical field by field even before this patch (I fail to see the difference) 🤔

In this order, fields are:
> u32, u32, u16, u16, u16, u16

Just thought it was curious, but this is looking good to me!

RRRadicalEdward:
@CBenoit The difference was in `proxy_server_offset` and `client_tls_cred_offset`. They point either to ANSI or a wide strings. We construct proxy_server and client_tls_cred strings from this values

PR #137, opened 2023-07-14
Fix NTLM downgrade with IP address

awakecoding:


CBenoit (approved):
Looks reasonable to me

PR #138, opened 2023-07-20
refactor(network_client): Refactor NetworkClient trait

pacmancoder:
Refactored `NetworkClient` API as required by https://github.com/Devolutions/IronRDP/issues/155#issuecomment-1593724348

- Added the `NetworkClient::supported_protocols` method to discover if the underlying network client implementation supports the given protocol
- Added the `NetworkClient::name` method to provide more context for the error messages
- Moved ASN.1 messages serialization/deserialization for `NetworkClient::send_http` from implementors to sspi-rs internal code.
- Removed `NetworkClient::send_http` and squashed it to single`NetworkClient::send` (now a signature for all tcp/udp/http/https methods is the same)

### DIFF
--- a/src/kerberos/mod.rs
+++ b/src/kerberos/mod.rs
@@ -14,9 +14,9 @@ use picky_asn1_x509::oids;
 use picky_krb::constants::gss_api::AUTHENTICATOR_CHECKSUM_TYPE;
 use picky_krb::constants::key_usages::ACCEPTOR_SIGN;
 use picky_krb::crypto::CipherSuite;
-use picky_krb::data_types::{KrbResult, ResultExt};
+use picky_krb::data_types::{KerberosStringAsn1, KrbResult, ResultExt};
 use picky_krb::gss_api::{NegTokenTarg1, WrapToken};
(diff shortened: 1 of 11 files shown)


src/network_client.rs
pacmancoder, line 1: GitHub is showing this diff in rather unreadable form, I suggest seeing it in side-to-side diff 😉

src/kerberos/mod.rs
```diff
@@ -153,16 +156,43 @@ impl Kerberos {
...
+                        self.config.network_client.supported_protocols(),
+                    ),
+                ));
+            }
+
+            return match protocol {
+                NetworkProtocol::Tcp | NetworkProtocol::Udp => {
+                    self.config.network_client.send(protocol, &kdc_url, data)
+                }
+                NetworkProtocol::Http | NetworkProtocol::Https => {
```
pacmancoder, line 182: Now `IronRDP` and other crates which depend on `sspi-rs` don't need to know about `KdcProxyMessage`, `picky`, etc.

awakecoding (approved):
LGTM

PR #140, opened 2023-08-07
fix!: invalid address when using UDP

CBenoit:
- Trim the scheme from the `url` parameter (UDP target address is something like `hostname:port`).
- Extract the message length exclusion logic so that user don’t need to handle it himself.
- Take ownership of the `url` field to remove unneeded copies.
- Bump version to 0.10.0


awakecoding (approved):

pacmancoder (approved):
LGTM!

PR #144, opened 2023-08-25
Kerberos smart card module

TheBestTvarynka:
Hi,
In this PR I've added a Kerberos smart card module. This module can communicate with the real smart card using the WinSCard API to sign some data.

You can notice that the code contains a lot of hardcoded values like input buffers for the `SCardTransmit` function, control code for the `SCardControl` function, etc. At the current moment, we do not have any APDU command builder or anything else. So, we are forced to use hardcoded values. I think, when I implement the second phase (iteration) we'll be able to reuse some code and refactor this module.

How I know what APDU command to use:
I recorded API calls and extracted needed APDUs for the data signing.

_Note_: It doesn't contain actual smart card logon implementation. I'll open a separate PR for it.

Docs & references:

* [`winscard.h` header](https://learn.microsoft.com/en-us/windows/win32/api/winscard)
* [`pcsc` crate](https://docs.rs/pcsc/latest/pcsc)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -94,6 +94,7 @@ mod kdc;
 mod krb;
 mod secret;
 #[cfg(feature = "scard")]
+#[allow(dead_code)]
 mod smartcard;
 mod utils;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -93,6 +93,8 @@ mod dns;
 mod kdc;
 mod krb;
 mod secret;
+#[cfg(feature = "scard")]
+mod smartcard;
 mod utils;
 
 #[cfg(all(feature = "tsssp", not(target_os = "windows")))]
--- a/src/smartcard.rs
+++ b/src/smartcard.rs
@@ -1,0 +1,163 @@
+use std::fmt;
+
+use pcsc::{Context, Scope, Protocols, ShareMode, Card};
+use picky_asn1::wrapper::OctetStringAsn1;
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -93,6 +93,9 @@ mod dns;
 mod kdc;
 mod krb;
 mod secret;
+#[cfg(feature = "scard")]
+#[allow(dead_code)]
```
TheBestTvarynka, line 97: it's a temporary code. just to pass the CI and avoid warnings about unused code

CBenoit: Sounds good :+1:

src/smartcard.rs
```diff
@@ -0,0 +1,171 @@
...
+        match &self.smart_card_type {
+            SmartCardApi::WinSCard(scard) => {
+                // https://www.eftlab.com/knowledge-base/complete-list-of-apdu-responses
+                const APDU_RESPONSE_OK: [u8; 2] = [0x90, 0x00];
+
+                // this control code is extracted from the API calls recording during the mstsc connection establishing
+                scard.control(0x00313520, &[], &mut [])?;
+
+                let mut result_buff = [0; 16];
+                #[rustfmt::skip]
```
TheBestTvarynka, line 58: `rustfmt` mixes values with comments and makes it less readable

CBenoit (approved):
LGTM!

PR #165, opened 2023-09-20
Fix UPN handling

### LINKED ISSUE #150: mstsc.exe unpacked UPN-formatted credentials trailing characters
from the SSPI logs, UPN-formatted credentials (no domain, everything in user field) have trailing characters:

image (49)

"DOMAIN\user" which gets split in two parts don't suffer from the same issue:
image (48)

The bug most likely happens right after the calls to CredUnPackAuthenticationBufferW in unpack_sec_winnt_auth_identity_ex2_w

TheBestTvarynka:
Hi,
I fixed the UPN (user principal name) handling. closes #150 

The problem was in the trailing null characters at the end of the username buffer. In the case of UPN logon, the username buffer has two null bytes at the end.

Also, I recorded a little video demonstration:

* logon using UPN

https://github.com/Devolutions/sspi-rs/assets/43034350/1d3f99be-4d99-4cc4-8a9d-d01f19ac969b

* logon using domain + username

https://github.com/Devolutions/sspi-rs/assets/43034350/ec4ef518-6fca-487b-a35f-1f14dd9aac35

cc @awakecoding

### DIFF
--- a/ffi/src/utils.rs
+++ b/ffi/src/utils.rs
@@ -11,17 +11,6 @@ pub fn into_raw_ptr<T>(value: T) -> *mut T {
     Box::into_raw(Box::new(value))
 }
 
-pub fn vec_into_raw_ptr<T>(v: Vec<T>) -> *mut T {
-    Box::into_raw(v.into_boxed_slice()) as *mut T
-}
-
-pub unsafe fn raw_w_str_to_bytes(raw_buffer: *const u16, len: usize) -> Vec<u8> {
-    from_raw_parts(raw_buffer, len)
-        .iter()
-        .flat_map(|w_char| w_char.to_le_bytes())
-        .collect()
-}
-
 pub unsafe fn c_w_str_to_string(s: *const SecWChar) -> String {
     let mut len = 0;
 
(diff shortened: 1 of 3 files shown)


ffi/src/utils.rs
```diff
@@ -50,3 +39,16 @@ pub unsafe fn transform_credentials_handle<'a>(
...
+pub fn raw_wide_str_trim_nulls(raw_str: &mut Vec<u8>) {
+    let mut len = raw_str.len();
+    while len > 2 {
+        if raw_str[len - 2..] == [0, 0] {
+            raw_str.truncate(len - 2);
+            len = raw_str.len();
+        } else {
+            break;
+        }
+    }
```
CBenoit, line 53: Maybe you could add a comment explaining that it works because when encoding an UTF-16 character using two code units, the 16-bit values are chosen from the UTF-16 surrogate range 0xD800–0xDFFF, and thus only `\0` is encoded by two consecutive null bytes.

CBenoit (approved):
Looks good to me

CBenoit (approved):

PR #166, opened 2023-09-20
Fix `execute_future` not working with `tokio::spawn`

probakowski:
In some cases (e.g. when using async IO, like TCP connection) resolving DNS queries for KDC will block when used in `tokio::spawn`-ed task. AFAICT this happens because we block thread that is used by IO and timers.
This change fixes that with use `block_in_place` so the thread we're on will be marked by tokio as blocked and other task will be moved to the new one.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ exclude = [
 [features]
 default = []
 network_client = ["dep:reqwest", "dep:portpicker"]
-dns_resolver = ["dep:trust-dns-resolver", "dep:tokio"]
+dns_resolver = ["dep:trust-dns-resolver", "dep:tokio", "tokio/rt-multi-thread"]
 # TSSSP should be used only on Windows as a native CREDSSP replacement
 tsssp = ["dep:rustls"]
 # Turns on Kerberos smart card login (available only on Windows and users WinSCard API)
@@ -63,7 +63,7 @@ num-bigint-dig = "0.8"
 tracing = "0.1"
 rustls = { version = "0.21", features = ["dangerous_configuration"], optional = true }
 zeroize = { version = "1.6", features = ["zeroize_derive"] }
-tokio = { version = "1.32", features = ["time", "rt", "rt-multi-thread"], optional = true }
+tokio = { version = "1.32", features = ["time", "rt"], optional = true }
 pcsc = { version = "2.8", optional = true }
 
 [target.'cfg(windows)'.dependencies]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -63,7 +63,7 @@ num-bigint-dig = "0.8"
 tracing = "0.1"
 rustls = { version = "0.21", features = ["dangerous_configuration"], optional = true }
 zeroize = { version = "1.6", features = ["zeroize_derive"] }
-tokio = { version = "1.32", features = ["time", "rt"], optional = true }
+tokio = { version = "1.32", features = ["time", "rt", "rt-multi-thread"], optional = true }
```
CBenoit, line 66: Do we really need to enable the "rt-multi-thread" feature for `tokio`?

probakowski: `tokio::block_in_place` is guarded by check for `rt-multi-thread` feature. We can either enable it here or add another feature in `sspi` to and add it there (with `cfg` guard in code)

CBenoit: Oh okay! In that case, I think it’s good as-is, let’s not add more complexity to this. Longer term, I also have a plan for removing the dependency to tokio completely and leave the user spawn a task in their runtime directly.

Cargo.toml
```diff
@@ -23,7 +23,7 @@ exclude = [
 [features]
 default = []
 network_client = ["dep:reqwest", "dep:portpicker"]
-dns_resolver = ["dep:trust-dns-resolver", "dep:tokio"]
+dns_resolver = ["dep:trust-dns-resolver", "dep:tokio", "tokio/rt-multi-thread"]
```
CBenoit, line 26: This change is good, thank you :+1:

Cargo.toml
```diff
@@ -75,7 +75,7 @@ windows-sys = { version = "0.48", features = ["Win32_Security_Cryptography", "Wi
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 async-dnssd = "0.5"
 futures = "0.3"
-tokio = { version = "1.32", features = ["time", "rt"] }
+tokio = { version = "1.32", features = ["time", "rt", "rt-multi-thread"] }
```
CBenoit, line 78: Since the feature is implied by `dns_resolver`, you can probably remove it from here too?

probakowski: `#[cfg(any(feature = "dns_resolver", target_os = "macos", target_os = "ios"))]`<-  you have to enable it for either `dns_resolver` or `macos`/`ios`, so it's required in both places

CBenoit: Oh, okay! :+1:

CBenoit (approved):
Thank you!

Issue #168, opened 2023-09-20
Suppoer QueryContextAttributes SECPKG_ATTR_SESSION_KEY


jborean93:
Would be great to support `SECPKG_ATTR_SESSION_KEY` with [QueryContextAttributesW](https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-querycontextattributesw) to retrieve the generated session key for the security context. This is needed if you ever want to implement support for SMB authentication as it uses this key to derive the subsequent keys it uses internally.

PR #175, opened 2023-09-24
Store and give access to session_key in Ntlm

bobbobbio:
I'm using this crate to add NTLM support to my SMB 3.1.1 client I'm developing.
The session_key is used as part of the calculation of the signing key. See this page for details https://learn.microsoft.com/en-us/archive/blogs/openspecification/smb-2-and-smb-3-security-in-windows-10-the-anatomy-of-signing-and-cryptographic-keys

This pull request stashes the session_key that was used in the Ntlm struct, and provides access to it with an accessor function.


jborean93:
Slightly related, this is a request to expose this same value through the normal Win32 APIs https://github.com/Devolutions/sspi-rs/issues/168.

CBenoit (approved):
LGTM, thank you!

PR #177, opened 2023-09-28
Add wasm-bindgen feature for tima crate

irvingoujAtDevolution:
add feature for Kerberos to work in browser

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,15 +11,8 @@ description = "A Rust implementation of the Security Support Provider Interface 
 keywords = ["ntlm", "auth", "sspi", "windows", "kerberos"]
 
 [workspace]
-members = [
-  "ffi",
-  "ffi/symbol-rename-macro",
-]
-exclude = [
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -11,15 +11,8 @@ description = "A Rust implementation of the Security Support Provider Interface
 keywords = ["ntlm", "auth", "sspi", "windows", "kerberos"]
 
 [workspace]
-members = [
```
irvingoujAtDevolution, line 14: most things are automatically formated

Cargo.toml
```diff
@@ -34,7 +27,10 @@ byteorder = "1.4"
 bitflags = "2.4"
 rand = "0.8"
 cfg-if = "1"
-time = { version = "0.3", default-features = false, features = ["std"] }
+time = { version = "0.3", default-features = false, features = [
+  "std",
+  "wasm-bindgen",
```
irvingoujAtDevolution, line 32: only this line is the real change

PR #178, opened 2023-10-06
[ARC-144] add if condition for HTTP target, allow winrm connection

irvingoujAtDevolution:
winrm connection seems disagree with Ker5 user to user protocol, add a new if condition, if the target is http (aka winrm server), then use plain kerbero5 protocol instead of user to user


src/kerberos/mod.rs
```diff
@@ -890,7 +1076,7 @@ impl<'a> Kerberos {
                 info!(encoded_ap_req_authenticator = ?encoded_auth);
 
                 // FIXME: properly negotiate mech id - Windows always does KRB5 U2U
-                let mech_id = oids::krb5_user_to_user();
+                let mech_id = oids::ms_krb5();
```
awakecoding, line 1079: can you keep the KRB5 U2U for connections that aren't WinRM? we know RDP always uses KRB5 U2U

.github/workflows/ci.yml
```diff
@@ -77,7 +77,7 @@ jobs:
       - uses: actions/checkout@v3
 
       - name: Test
-        run: cargo test --manifest-path ${{ matrix.manifest }} ${{ matrix.additional-args }}
+        run: cargo test --manifest-path ${{ matrix.manifest }} ${{ matrix.additional-args }} --test *
```
irvingoujAtDevolution, line 80: this will avoid test examples, as examples/kerberos.rs require a real KDC and server to work

CBenoit: You probably meant `--tests` (with a final `s`), but this is NOT required. Examples an never _run_ as tests, they are only checked for compilation errors. You can read more here: https://doc.rust-lang.org/cargo/commands/cargo-test.html#target-selection
Note that even if we specified `--exemples`, it would only run _unit tests_ defined in example code, and that would be desirable.

src/kerberos/mod.rs
```diff
@@ -653,40 +654,225 @@ impl<'a> Kerberos {
...
+        };
+        self.realm = Some(realm.clone());
+
+        let options = GenerateAsReqOptions {
+            realm: &realm,
+            username: &username,
+            cname_type,
+            snames: &[TGT_SERVICE_NAME, &realm],
+            // 4 = size of u32
+            nonce: &OsRng.gen::<[u8; 4]>(),
```
CBenoit, line 726: Note that you could use `core::mem::size_of::<u32>()` so the code is self-explanatory

src/kerberos/mod.rs
```diff
@@ -653,40 +654,225 @@ impl<'a> Kerberos {
...
+
+        let as_rep = self.as_exchange(yield_point, &kdc_req_body, pa_data_options).await?;
+
+        info!("AS exchange finished successfully.");
+
+        self.realm = Some(as_rep.0.crealm.0.to_string());
+
+        let (encryption_type, salt) = extract_encryption_params_from_as_rep(&as_rep)?;
+
+        let encryption_type = CipherSuite::try_from(encryption_type as usize)?;
```
CBenoit, line 754: Use `usize::from` instead of `as` so we’re sure the conversion is lossless (`From` trait is for lossless conversions only). In practice this prevents conversions from turning into silent lossy conversions if the types of the input expressions ever change, and it’s also easier for people reading the code to know that the conversion is lossless.

```suggestion
        let encryption_type = CipherSuite::try_from(usize::from(encryption_type))?;
```

src/kerberos/mod.rs
```diff
@@ -653,40 +654,225 @@ impl<'a> Kerberos {
...
+            self.encryption_params.session_key.as_ref().unwrap(),
+            &authenticator,
+            &self.encryption_params,
+            context_requirements.into(),
+        )?;
+
+        let generated_neg_ap_req = generate_neg_ap_req_plain(ap_req, mech_id)?;
+        let encoded_neg_ap_req = picky_asn1_der::to_vec(&generated_neg_ap_req)?;
+        let output_token = SecurityBuffer::find_buffer_mut(builder.output, SecurityBufferType::Token)?;
+        // the out put token is already a certificate instead of a contentInfo here, need to figure out why
```
CBenoit, line 851: Maybe add a `TODO` comment, and create a ticket if this is an important detail that must be tracked.
```suggestion
        // TODO: the out put token is already a certificate instead of a contentInfo here, need to figure out why
```

CBenoit (commented):
Thank you!

.github/workflows/ci.yml
irvingoujAtDevolution: updated, the main problem was CI check complains about the feature used in the kerberos example, updated the the example, now it's guarded by cfg feature attribute.

Cargo.toml
```diff
@@ -82,3 +95,7 @@ tokio = { version = "1.32", features = ["time", "rt"] }
 static_assertions = "1"
 whoami = "1.4"
 picky = { version = "7.0.0-rc.8", features = ["x509"] }
+base64 = "0.21.4"
+hyper = "0.14.27"
+tracing-log = "0.1.3"
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated

examples/kerberos.rs
```diff
@@ -0,0 +1,140 @@
...
+};
+use sspi::{Kerberos, SspiImpl};
+use std::error::Error;
+use tracing::debug;
+use tracing_log::log::info;
+const TARGET_NAME: &'static str = "HTTP/your-server-name.your.domain";
+
+fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
+    tracing_subscriber::fmt()
+        .with_max_level(tracing::Level::DEBUG) // Adjust level as needed
```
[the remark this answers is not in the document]
irvingoujAtDevolution: udpated

Cargo.toml
```diff
@@ -61,17 +59,32 @@ trust-dns-resolver = { version = "0.23", optional = true }
 portpicker = { version = "0.1", optional = true }
 num-bigint-dig = "0.8"
 tracing = "0.1"
-rustls = { version = "0.21", features = ["dangerous_configuration"], optional = true }
+rustls = { version = "0.21", features = [
+  "dangerous_configuration",
+], optional = true }
```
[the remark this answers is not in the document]
irvingoujAtDevolution: updated, both the my disabled my vscode formatter and the file. Do you have a recommanded toml formatter tool for this?

src/kerberos/mod.rs
irvingoujAtDevolution: this is direct copy from original code

examples/kerberos.rs
```diff
@@ -53,89 +39,116 @@ fn main() -> Result<(), Box<dyn Error + Send + Sync>> {
...
 }
 
-fn get_cred_handle(kerberos: &mut Kerberos) -> AcquireCredentialsHandleResult<Option<CredentialsBuffers>> {
-    let username = "user@your.domain".to_string();
-    let password = "user's_password".to_string();
-    let identity = sspi::AuthIdentity {
-        username,
-        password: password.into(),
-        domain: None,
+#[cfg(feature = "network_client")]
```
CBenoit, line 42: You can put this at the top of the file:
```
#![cfg(feature = "network_client")]
```

src/kerberos/mod.rs
CBenoit: Oh okay, I understand. Do you think you could address this anyway while you’re at it? Thank you

examples/kerberos.rs
irvingoujAtDevolution: the problem is that ```cargo test``` needs to see the main function present in the file. if I put it on top, it will not pass ```cargo test```

CBenoit: You’re right! Actually, in this case you should remove the`#[cfg]`s altogether, and add this at the bottom of the `Cargo.toml` instead:
```rust
[[example]]
name = "kerberos"
required-features = ["network_client"]
```

This specifies that this example needs the "network_client" feature enabled in order to be compiled and it will be ignored if the feature is not available.

examples/kerberos.rs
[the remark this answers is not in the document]
CBenoit: Marking as unresolved as you do not appear to be using `EnvFilter` yet

irvingoujAtDevolution:
Not correct path, problem solved without using the newly implemented variation of Kerb5 protocol

PR #179, opened 2023-10-17
A tiny fix to Kerberos inner initiate sercurity context

irvingoujAtDevolution:


src/kerberos/mod.rs
```diff
@@ -653,7 +653,7 @@ impl<'a> Kerberos {
         Ok(())
     }
 
-    pub async fn initialize_security_context_impl(
+    pub(crate) async fn initialize_security_context_impl(
```
CBenoit, line 656: Can you elaborate what is fixed by this, please?

PR #180, opened 2023-10-18
fix: small fix for generator related changes and variable naming

irvingoujAtDevolution:
Updated Kerberos config debug print, renamed pku2u hostname to client_hostname for clarity. Updated Kerbero inner "private" initialize security context to Pub(crate)


src/kerberos/mod.rs
```diff
@@ -653,7 +653,7 @@ impl<'a> Kerberos {
         Ok(())
     }
 
-    pub async fn initialize_security_context_impl(
+    pub(crate) async fn initialize_security_context_impl(
```
irvingoujAtDevolution, line 656: stay consistent with NTLM, pku2u, make it easier to call with only
```
kerberos.initialize_security_context_impl(&mut builder)
```
as shown in example kerberos.rs

examples/kerberos.rs
CBenoit, line 1: Can you review the suggestions I made for the example in this PR, please?
https://github.com/Devolutions/sspi-rs/pull/178

irvingoujAtDevolution: updated, let me know if there's anything else should be fix as well.

examples/kerberos.rs
```diff
@@ -0,0 +1,150 @@
...
+use sspi::{Kerberos, SspiImpl};
+use std::error::Error;
+
+static HOSTNAME: &'static str = "computer_name.domain";
+static USERNAME: &'static str = "user@domain";
+static PASSWORD: &'static str = "Passoword";
+static AUTH_METHOD: &'static str = "Negotiate"; // Negotiate or Kerberos
+
+fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+    let kdc_url = std::env::var("SSPI_KDC_URL").expect("a KDC URL set in SSPI_KDC_URL");
```
irvingoujAtDevolution, line 20: use env var for kdc url, not sure if we should use env var for other variables as well. Please let me know and I can update accordingly. Thanks.

src/kerberos/config.rs
```diff
@@ -7,19 +7,12 @@ use crate::kdc::detect_kdc_url;
 use crate::negotiate::{NegotiatedProtocol, ProtocolConfig};
 use crate::{Kerberos, Result};
 
+#[derive(Debug)]
```
irvingoujAtDevolution, line 10: We can now derive Debug directly, as we no longer injects network client

src/pku2u/config.rs
```diff
@@ -9,28 +9,28 @@ use crate::{Pku2u, Result};
 pub struct Pku2uConfig {
     pub p2p_certificate: Certificate,
     pub private_key: SecretPrivateKey,
-    pub hostname: String,
+    pub client_hostname: String,
```
irvingoujAtDevolution, line 12: renaming to avoid miss understanding in the future

examples/kerberos.rs
```diff
@@ -0,0 +1,150 @@
...
+        Ok(result) => {
+            let output_buffer = secure_output_buffer[0].to_owned();
+            let res = (
+                base64::engine::general_purpose::STANDARD.encode(output_buffer.buffer),
+                result.status,
+            );
+            res
+        }
+        Err(_) => {
+            panic!("error steping");
```
CBenoit, line 147:
```suggestion
            panic!("error stepping");
```

examples/kerberos.rs
CBenoit: Thank you. Can you address the remaining as well?
https://github.com/Devolutions/sspi-rs/pull/178#discussion_r1349323842
https://github.com/Devolutions/sspi-rs/pull/178#discussion_r1349317663

examples/kerberos.rs
CBenoit: Yes please, it should not be required to modify the source code in order to run the example.
https://github.com/Devolutions/sspi-rs/pull/178#discussion_r1349321226
https://github.com/Devolutions/sspi-rs/pull/178#discussion_r1349321721

examples/kerberos.rs
irvingoujAtDevolution: updated, except for the tracing_subscriber one, as in this branch, I did not use tracing_subscriber

CBenoit: How are you logging?

irvingoujAtDevolution: I did not log anything in this case, previous PR was for debugging (as it was not working before). Since it is an working example now, for users who want to know what's going on, they may use any logging methods/debuggers they prefer. If you think it is necessary to log information in the example, I can put the tracing subscriber back.

CBenoit: Yes, this would be useful. Do something like this, but log to the terminal instead: https://github.com/Devolutions/sspi-rs/blob/949d36008baa37dd9930a0c378f90b72297071db/ffi/src/logging.rs#L34-L37
This way, the user can modify the log directives using the `SSPI_LOG_LEVEL` env variable, without having to modify the source code himself (future self will be happy).

irvingoujAtDevolution: updated

CBenoit: Perfect, thank you!

Cargo.toml
```diff
@@ -87,3 +87,4 @@ base64 = "0.21.4"
 static_assertions = "1"
 whoami = "1.4"
 picky = { version = "7.0.0-rc.8", features = ["x509"] }
+tracing-subscriber = "0.3.17"
```
CBenoit, line 90:
```suggestion
tracing-subscriber = "0.3"
```

Cargo.toml
```diff
@@ -79,6 +83,8 @@ futures = "0.3"
 tokio = { version = "1.32", features = ["time", "rt"] }
 
 [dev-dependencies]
+base64 = "0.21.4"
```
CBenoit, line 86:
```suggestion
base64 = "0.21"
```

CBenoit (approved):
LGTM! Thank you!

PR #182, opened 2023-10-26
Ntlm fixes for disabled sealing and signing

pauldumais:
1) NTLM flags now allow to disable signing and sealing
2) MIC is now properly generated if key exchange is disabled (such as the case of disabled sealing)
3) NTLM DecryptMessage now supports STREAM decoding

### DIFF
--- a/src/ntlm/messages/client/authenticate.rs
+++ b/src/ntlm/messages/client/authenticate.rs
@@ -123,7 +123,7 @@ pub fn write_authenticate(
     let mut encrypted_session_key = [0x00; ENCRYPTED_RANDOM_SESSION_KEY_SIZE];
     encrypted_session_key.clone_from_slice(encrypted_session_key_vec.as_ref());
 
-    context.flags = get_flags(&context, credentials);
+    context.flags = get_flags(context, credentials);
 
     if !context.flags.contains(NegotiateFlags::NTLM_SSP_NEGOTIATE_KEY_EXCH) {
         session_key = key_exchange_key;
--- a/src/ntlm/messages/client/authenticate.rs
+++ b/src/ntlm/messages/client/authenticate.rs
@@ -118,16 +118,17 @@ pub fn write_authenticate(
         ntlm_v2_hash.as_ref(),
         challenge_message.timestamp,
     )?;
-    let mut session_key = OsRng.gen::<[u8; SESSION_KEY_SIZE]>();
+    context.flags = get_flags(context, credentials);
+
+    let session_key = if context.flags.contains(NegotiateFlags::NTLM_SSP_NEGOTIATE_KEY_EXCH) {
+        OsRng.gen::<[u8; SESSION_KEY_SIZE]>()
+    } else {
+        key_exchange_key
+    };
+
     let encrypted_session_key_vec = Rc4::new(&key_exchange_key).process(session_key.as_ref());
     let mut encrypted_session_key = [0x00; ENCRYPTED_RANDOM_SESSION_KEY_SIZE];
     encrypted_session_key.clone_from_slice(encrypted_session_key_vec.as_ref());
-
-    context.flags = get_flags(context, credentials);
-
-    if !context.flags.contains(NegotiateFlags::NTLM_SSP_NEGOTIATE_KEY_EXCH) {
-        session_key = key_exchange_key;
-    }
 
     let message_fields = AuthenticateMessageFields::new(
         credentials,
--- a/src/ntlm/messages/client/authenticate.rs
+++ b/src/ntlm/messages/client/authenticate.rs
@@ -118,12 +118,17 @@ pub fn write_authenticate(
         ntlm_v2_hash.as_ref(),
         challenge_message.timestamp,
     )?;
-    let session_key = OsRng.gen::<[u8; SESSION_KEY_SIZE]>();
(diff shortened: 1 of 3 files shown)


src/ntlm/messages/client/authenticate.rs
```diff
@@ -118,12 +118,17 @@ pub fn write_authenticate(
...
     let encrypted_session_key_vec = Rc4::new(&key_exchange_key).process(session_key.as_ref());
     let mut encrypted_session_key = [0x00; ENCRYPTED_RANDOM_SESSION_KEY_SIZE];
     encrypted_session_key.clone_from_slice(encrypted_session_key_vec.as_ref());
 
-    context.flags = get_flags(context.flags, credentials);
+    context.flags = get_flags(context, credentials);
+
+    if !context.flags.contains(NegotiateFlags::NTLM_SSP_NEGOTIATE_KEY_EXCH) {
+        session_key = key_exchange_key;
+    }
```
CBenoit, line 130: Nit: I understand that we re-use `key_exchange_key` unless the `NTLM_SSP_NEGOTIATE_KEY_EXCH` flag is set, in which case we generate a different session key. Instead of generating a session key that we never use and use a mutable binding, you could do that:
```rust
    let session_key = if context.flags.contains(NegotiateFlags::NTLM_SSP_NEGOTIATE_KEY_EXCH) {
        OsRng.gen::<[u8; SESSION_KEY_SIZE]>()
    } else {
        key_exchange_key
    };
```
(if blocks are expressions, and this is more idiomatic)

src/ntlm/mod.rs
```diff
@@ -398,10 +416,20 @@ impl Sspi for Ntlm {
...
+            SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Stream)?
+        }
+        .buffer
+        .clone();
         let data = SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Data)?;
+        encrypted.extend_from_slice(&data.buffer);
 
-        *data.buffer.as_mut() = self.recv_sealing_key.as_mut().unwrap().process(data.buffer.as_slice());
+        let signature = encrypted[0..16].to_vec();
+        let enc = encrypted[16..].to_vec();
```
CBenoit, line 430: Question: do you know if the "invariant" `buffer.len() > 16` is checked somewhere? Inside `SecurityBuffer` perhaps? Just being cautious, because in Rust, when the index is out of bounds, such indexing, instead of corrupting the memory, will stop the program by panicking (similar to C++ stack unwinding when an exception is raised). However, panicking (or unwinding) across an FFI boundary is undefined behavior. (There is [a working group on that matter](https://blog.rust-lang.org/inside-rust/2020/02/27/ffi-unwind-design-meeting.html), but as of today it’s undefined.) FYI, the safe route is to use `.get`:
```rust
let signature = encrypted.get(0..16)?;
let enc = &encrypted[16..]; // the second one is fine, because of the above
```
You may have do something like `get(0..16).or_else(|| …)` in order to convert the `Option` into a `Result`.
Alternatively, do the usual business with something like `if encrypted.len() < 16`.

CBenoit, line 430: Nit: as far as I can see in the rest of the code, there is no need to use `.to_vec()`:
```suggestion
        let signature = &encrypted[0..16];
        let enc = &encrypted[16..];
```
(and remove the `.as_slice()` below)

src/ntlm/mod.rs
```diff
@@ -398,10 +416,20 @@ impl Sspi for Ntlm {
...
         }
 
-        SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Token)?; // check if exists
+        let mut encrypted = if let Ok(buffer) = SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Token) {
+            buffer
+        } else {
+            SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Stream)?
+        }
+        .buffer
+        .clone();
```
CBenoit, line 425: Nit: as far as I know, it’s more idiomatic to write this:
```suggestion
        let security_buffer = let Ok(buffer) = SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Token) {
            buffer
        } else {
            SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Stream)?
        };
        let encrypted = &security_buffer.buffer;
```
If I understand correctly, it’s also not necessary to `.clone()` the buffer at all.

CBenoit (commented):
I don’t understand most of it, but I left a few comments FYI. Feel free to ignore the nits.

RRRadicalEdward (approved):
Benoît has written all the comments. I have nothing to add

src/ntlm/messages/client/authenticate.rs
pauldumais: Ok

src/ntlm/mod.rs
pauldumais: Done

awakecoding:
the changes don't look like they'd break other ways to call the same APIs, so I'm good with them. I'll leave the rest for @CBenoit to review in terms of Rust code style :)

awakecoding (approved):

src/ntlm/mod.rs
pauldumais: This code was copied from the kerberos code, and the changes proposed do not compile, I will leave it as is since it's only a nit.

PR #186, opened 2023-11-10
Tiny Updates For Kerbero Example 

irvingoujAtDevolution:
Fix compile error for Kerberos  example


CBenoit (approved):
Oh thank you! This is strange the CI didn't catch this problem

PR #187, opened 2023-11-10
Fix: NTLM not working in IronRDP, Add Flag to enforce sign and seal

irvingoujAtDevolution:
This PR is currently for discussion only. we could hard code the flag in CredSSP, but might not be desired. 

This fixes problem in [IronRDP PR, Enable Kerberos](https://github.com/Devolutions/IronRDP/pull/260).

The changes will force CredSSP to sign and seal the message, which allows it to have the same behavior as before


awakecoding:
CredSSP always does signing and sealing, so this fix looks good to me

pauldumais (approved):
LGTM

PR #189, opened 2023-12-12
Fix: the GSS Flags were hard coded, which does not work for Ldaps,Fixed.

irvingoujAtDevolution:
Note: This change might be breaking. Confidentiality and integrity is not implied anymore. User needs to actively specify these two flag to get sign and seal. 

This change is critical for LDAPs, as Active Directory does not accept Kerberos with sign and seal flags.

This implementation requires reviewers attention, in particular, at line 408 of src/kerberos/client/generators.rs. Where windows never specified how these flags map to the GSS checksum flags. This is purely guessed from their[ Kerberos documentation](https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-KILE/%5bMS-KILE%5d.pdf) at section 3.2.5.2. and the [SSPI documentation ](https://learn.microsoft.com/en-us/windows/win32/secauthn/context-requirements). 

Whether this implementation is correct or not, we need a way for caller to change that flag.

### DIFF
--- a/src/kerberos/client/generators.rs
+++ b/src/kerberos/client/generators.rs
@@ -362,9 +362,9 @@ impl Default for ChecksumValues {
     }
 }
 
-impl Into<Vec<u8>> for ChecksumValues {
-    fn into(self) -> Vec<u8> {
-        self.inner
+impl From<ChecksumValues> for Vec<u8> {
+    fn from(val: ChecksumValues) -> Self {
+        val.inner
     }
 }
 
--- a/src/kerberos/client/generators.rs
+++ b/src/kerberos/client/generators.rs
@@ -1,5 +1,6 @@
 use std::str::FromStr;
 
+use bitflags;
 use md5::{Digest, Md5};
 use picky_asn1::bit_string::BitString;
 use picky_asn1::date::GeneralizedTime;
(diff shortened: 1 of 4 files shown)


awakecoding:
It looks good to me, I'm surprised it has worked so far. We'll need to test that it doesn't break things for other protocols though

awakecoding (approved):

src/kerberos/client/generators.rs
```diff
@@ -345,7 +346,100 @@ pub fn generate_tgs_req(options: GenerateTgsReqOptions) -> Result<TgsReq> {
...
+        self.inner[20..24].copy_from_slice(&flag_bytes);
+    }
+
+    pub(crate) fn get_inner(self) -> Vec<u8> {
+        self.inner
+    }
+}
+
+bitflags::bitflags! {
+    //https://datatracker.ietf.org/doc/html/rfc4121#section-4.1.1.1, this is crutial for LDAPS
```
CBenoit, line 392: typo:
```suggestion
    // https://datatracker.ietf.org/doc/html/rfc4121#section-4.1.1.1, this is crucial for LDAPS
```

src/kerberos/client/generators.rs
```diff
@@ -345,7 +346,100 @@ pub fn generate_tgs_req(options: GenerateTgsReqOptions) -> Result<TgsReq> {
 #[derive(Debug)]
 pub struct ChecksumOptions {
     pub checksum_type: Vec<u8>,
-    pub checksum_value: Vec<u8>,
+    pub checksum_value: ChecksumValues,
+}
+
+#[derive(Debug)]
+pub struct ChecksumValues {
+    inner: Vec<u8>, // use named fields for future extensibility, this is a temporary solution
```
CBenoit, line 354: suggestion: Use `[u8; 24]` instead of `Vec<u8>`.

```diff
diff --git a/src/kerberos/client/generators.rs b/src/kerberos/client/generators.rs
index d1b31df..cfcb1b7 100644
--- a/src/kerberos/client/generators.rs
+++ b/src/kerberos/client/generators.rs
@@ -351,18 +351,18 @@ pub struct ChecksumOptions {
 
 #[derive(Debug)]
 pub struct ChecksumValues {
-    inner: Vec<u8>, // use named fields for future extensibility, this is a temporary solution
+    inner: [u8; 24], // use named fields for future extensibility, this is a temporary solution
 }
 
 impl Default for ChecksumValues {
     fn default() -> Self {
         Self {
-            inner: AUTHENTICATOR_DEFAULT_CHECKSUM.to_vec(),
+            inner: AUTHENTICATOR_DEFAULT_CHECKSUM,
         }
     }
 }
 
-impl From<ChecksumValues> for Vec<u8> {
+impl From<ChecksumValues> for [u8; 24] {
     fn from(val: ChecksumValues) -> Self {
         val.inner
     }
@@ -370,9 +370,7 @@ impl From<ChecksumValues> for Vec<u8> {
 
 impl From<[u8; 24]> for ChecksumValues {
     fn from(bytes: [u8; 24]) -> Self {
-        ChecksumValues {
-            inner: Vec::from(bytes),
-        }
+        ChecksumValues { inner: bytes }
     }
 }
 
@@ -383,7 +381,7 @@ impl ChecksumValues {
         self.inner[20..24].copy_from_slice(&flag_bytes);
     }
 
-    pub(crate) fn get_inner(self) -> Vec<u8> {
+    pub(crate) fn get_inner(self) -> [u8; 24] {
         self.inner
     }
 }
@@ -511,7 +509,7 @@ pub fn generate_authenticator(options: GenerateAuthenticatorOptions) -> Result<A
         }
         Optional::from(Some(ExplicitContextTag3::from(Checksum {
             cksumtype: ExplicitContextTag0::from(IntegerAsn1::from(checksum_type)),
-            checksum: ExplicitContextTag1::from(OctetStringAsn1::from(checksum_value)),
+            checksum: ExplicitContextTag1::from(OctetStringAsn1::from(checksum_value.to_vec())),
         })))
     } else {
         Optional::from(None)
diff --git a/src/pku2u/generators.rs b/src/pku2u/generators.rs
index a0113b0..ea9864b 100644
--- a/src/pku2u/generators.rs
+++ b/src/pku2u/generators.rs
@@ -254,7 +254,7 @@ pub fn generate_authenticator(options: GenerateAuthenticatorOptions) -> Result<A
         checksum_value,
     }) = checksum
     {
-        let mut checksum_value = checksum_value.get_inner();
+        let mut checksum_value = checksum_value.get_inner().to_vec();
         if checksum_type == AUTHENTICATOR_CHECKSUM_TYPE && channel_bindings.is_some() {
             if checksum_value.len() < 20 {
                 return Err(Error::new(
```

rationale: 1) Here, a checksum is always 24 bytes. Therefore, `[u8; 24]` is just the right type. 2) `Vec<u8>` is itself taking 24 bytes, excluding the 24 bytes of actual payload on the heap. `[u8; 24]` is 24 bytes by itself, payload only, no allocation. ⇒ There is no benefit in using `Vec<u8>`

src/kerberos/client/generators.rs
```diff
@@ -345,7 +346,100 @@ pub fn generate_tgs_req(options: GenerateTgsReqOptions) -> Result<TgsReq> {
...
+}
+
+impl ChecksumValues {
+    pub(crate) fn set_flags(&mut self, flags: GssFlags) {
+        let flag_bits = flags.bits();
+        let flag_bytes = flag_bits.to_le_bytes();
+        self.inner[20..24].copy_from_slice(&flag_bytes);
+    }
+
+    pub(crate) fn get_inner(self) -> Vec<u8> {
```
CBenoit, line 386: suggestion: Rename to `into_inner`.

rationale: The convention for methods deconstructing data into its underlying representation is `into_*`. [Rust API guidelines – C-CONV](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv).

CBenoit (approved):
Thank you for that! I left a few comments for code improvements, but I didn’t spot any significant issue just by looking at the Rust code. Feel free to address the comment (or not), and merge yourself once ready.

PS: I’ll rename the PR title to reflect the [conventional commit specification](https://www.conventionalcommits.org/en/v1.0.0/).

RRRadicalEdward (approved):

src/kerberos/client/generators.rs
irvingoujAtDevolution: The reason why I use Vec<u8> here is because it is actually not fixed to 24 bytes. According to the RFC 4121, 4.1.1, there are optional values that could be added in the future.

irvingoujAtDevolution:
@awakecoding , one thing to notice is that, this flag should also has an effect when encrypt/decrypt payloads. For example, if sign is on and seal is off, the message should comes with a signature but with unencrypted payload. I'll open a issue to track it

irvingoujAtDevolution:
image

here is a screenshot that it works no sign, no seal, plain TCP connection

image
here is another that works with no sign, no seal, tls connection but decrypted using @awakecoding TLS decryption script

src/kerberos/client/generators.rs
irvingoujAtDevolution: updated

src/kerberos/client/generators.rs
```diff
@@ -345,7 +346,100 @@ pub fn generate_tgs_req(options: GenerateTgsReqOptions) -> Result<TgsReq> {
...
+        self.inner[20..24].copy_from_slice(&flag_bytes);
+    }
+
+    pub(crate) fn into_inner(self) -> Vec<u8> {
+        self.inner
+    }
+}
+
+bitflags::bitflags! {
+    // https://datatracker.ietf.org/doc/html/rfc4121#section-4.1.1.1, this is crutial for LDAPS
```
CBenoit, line 392: typo:
```suggestion
    // https://datatracker.ietf.org/doc/html/rfc4121#section-4.1.1.1, this is crucial for LDAPS
```

TheBestTvarynka (approved):
LGTM 👍

jborean93:
> Note: This change might be breaking. Confidentiality and integrity is not implied anymore. User needs to actively specify these two flag to get sign and seal.

Don't you always want confidentiality and integrity by default, you should only not be setting it if the NO_INTEGRITY flag is specified.

jborean93:
Only when using `GSS-SPNEGO` because it uses the negotiated context flags to determine if signing/sealing was applied by the auth context. The `GSSAPI` SASL is fine with this as it negotiates the signing/sealed after the auth with another round trip of tokens. My comment was around not always having integrity and confidentiality by default and opting out of it with the `NO_INTEGRITY` flag. I.e. both SSPI and GSSAPI always negotiate integrity and confidentiality with Kerberos unless the relevant `NO_INTEGIRTY` flag is specified.

pauldumais:
Yes agreed we will make sure to fix this.

irvingoujAtDevolution:
@jborean93 added new issue #198 according to the thread.

PR #192, opened 2023-12-18
Fix Negotiate failure to fallback to NTLM

irvingoujAtDevolution:
When the first time InitializeSercurityContext is called for negotiate, the variable `self.auth_identity` is always None, but latter used when switching to NTLM or PKU2U, which of course will not work since it is None. Fixed the problem so Negotiate could properly fallback to NTLM with authentication information from the builder.


pauldumais (approved):
Tested it and it works for me.

CBenoit (approved):
Looks like a genuine overlook. Thank you!

PR #193, opened 2023-12-18
Expose RustSspiSetupLogger

irvingoujAtDevolution:
Forget to add it to sspi.def before, now correctly expose it to the caller.


CBenoit (approved):
LGTM

PR #194, opened 2023-12-18
Update README.md

irvingoujAtDevolution:
Readme is a bit outdated, I was updating my Visual Studio and my Rustc is not working so I figured to update this


CBenoit (approved):
Thank you!

PR #202, opened 2024-01-29
Restructure FFI

TheBestTvarynka:
Hi,
In this pull request, I've restructured the FFI crate and moved common bindings to the `crates/ffi-types` crate.

We plan to export SSPI and WinSCard functions from the one dll. So, I moved all SSPI-related bindings to the `sspi` module and placed the winscard functions definitions into the `winscard` module.

Also, I placed all winscard-related typings into a `ffi-types` crate. It'll help keep our code cleaner and reuse them when needed. Ideally, I think we should do the same with sspi: move sspi-related typing to this crate too.

Most of the changes are just typings. So, don't be afraid of this big number 😅 

Docs & references:
* [`winscard.h`](https://learn.microsoft.com/en-us/windows/win32/api/winscard/)


CBenoit (approved):
LGTM! Thank you!

TheBestTvarynka:
Finally, the CI is green

CBenoit:
Thank you! This time I’ll squash because there are intermediate commits which don’t pass the CI, so it’s not convenient for bisecting.

PR #203, opened 2024-01-31
WinScard and WinScardContex traits and their implementations

TheBestTvarynka:
Hi,
In this PR I've added the `WinScard` and `WinScardContext` traits and implemented them for `Scard` and `ScardContext`.
We need these features for better interaction between FFI bindings and our scard implementation.

Docs & references:
* [`winscard.h`](https://learn.microsoft.com/en-us/windows/win32/api/winscard/)


crates/winscard-rs/src/lib.rs
CBenoit, line 1: Just noticed that the folder is called `winscard-rs` instead of `winscard`. Do you think you could change this in another PR?

CBenoit (approved):
Well documented code! Thank you!

crates/winscard-rs/src/lib.rs
TheBestTvarynka: yes, I can change it

PR #204, opened 2024-02-01
WinSCard: refactoring and fixes

TheBestTvarynka:
Hi,
This pull request contains fixes to all the issues I faced during the debugging. Also, I refactored the code. So, now the `winscard` crate is fully ready.

These fixes contain a lot of tricky code. So I added comments to clarify this.
Also, I renamed the crate as you asked in [my previous PR](https://github.com/Devolutions/sspi-rs/pull/203#discussion_r1472660419).

Docs & references:
* [`winscard.h`](https://learn.microsoft.com/en-us/windows/win32/api/winscard/).
* [NIST.SP.800-73-4](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf).


crates/winscard-rs/src/lib.rs
TheBestTvarynka, line 1: I used the `git mv` command to rename the crate, but `git` still doesn't seem to be able to fully track such changes

CBenoit: I think that’s because some files were changed "too much" to be considered "moved". It’s okay!

CBenoit (approved):
LGTM!

PR #205, opened 2024-02-02
WinSCard FFI implementation

TheBestTvarynka:
Hi,
This pull request contains the WinSCard FFI implementation.

The WinSCard FFI is very tricky because Microsoft designed this API with complications. For example, during the smart card context deletion, we also need to free all allocated resources by this context (like smart card handles and buffers). For every dangerous/critical place, I added a comment about it.

I can't make this PR smaller 😞 . The smart card context and smart card handles are very coupled together, so I can't split this PR into two.

Docs & references:
* [`winscard.h`](https://learn.microsoft.com/en-us/windows/win32/api/winscard/).
* [NIST.SP.800-73-4](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf).


crates/ffi-types/src/winscard/functions.rs
```diff
@@ -8,23 +8,26 @@ use crate::{
...
 pub type SCardListInterfacesAFn = extern "system" fn(ScardContext, LpCStr, LpGuid, LpDword) -> ScardStatus;
 pub type SCardListInterfacesWFn = extern "system" fn(ScardContext, LpCWStr, LpGuid, LpDword) -> ScardStatus;
 pub type SCardGetProviderIdAFn = extern "system" fn(ScardContext, LpCStr, LpGuid) -> ScardStatus;
 pub type SCardGetProviderIdWFn = extern "system" fn(ScardContext, LpCWStr, LpGuid) -> ScardStatus;
 pub type SCardGetCardTypeProviderNameAFn =
-    extern "system" fn(ScardContext, LpCStr, u32, *mut u8, LpDword) -> ScardStatus;
+    unsafe extern "system" fn(ScardContext, LpCStr, u32, *mut u8, LpDword) -> ScardStatus;
 pub type SCardGetCardTypeProviderNameWFn =
-    extern "system" fn(ScardContext, LpCWStr, u32, *mut u16, LpDword) -> ScardStatus;
+    unsafe extern "system" fn(ScardContext, LpCWStr, u32, *mut u16, LpDword) -> ScardStatus;
```
CBenoit, line 30: question: What is the reason for adding the `unsafe` keyword here? IIRC, calling an external function from Rust is always unsafe, and I wasn’t aware it was possible at all to add the keyword. Does it change the semantic, or it’s just aesthetic? I suggest we keep a consistent style if there is no special semantic.

ffi/Cargo.toml
```diff
@@ -15,19 +15,29 @@ crate-type = ["cdylib"]
...
 
 # logging
 tracing = { version = "0.1" }
-tracing-subscriber = { version = "0.3", features = ["std", "fmt", "local-time", "env-filter"] }
+tracing-subscriber = { version = "0.3", features = [
+    "std",
+    "fmt",
+    "local-time",
+    "env-filter",
+] }
```
CBenoit, line 40: style (non-blocking): Did you use an automated formatting tool? I think this is not a very standard way of formatting the Cargo.toml, so there is no need to change to this on purpose, but it’s not very important.

ffi/src/winscard/scard_handle.rs
```diff
@@ -0,0 +1,183 @@
...
+use std::slice::{from_raw_parts, from_raw_parts_mut};
+
+use ffi_types::winscard::{LpScardIoRequest, ScardContext, ScardHandle, ScardIoRequest};
+use ffi_types::LpCVoid;
+use winscard::winscard::{IoRequest, Protocol, WinScard, WinScardContext};
+use winscard::{Error, ErrorKind, WinScardResult};
+
+/// Represents smart card context handle.
+/// Additionally, it holds allocated buffers and created smart card handles.
+/// We need them because during the smart card context deletion, we need to free all allcated resources.
```
CBenoit, line 11: nitpick: All the markdown until the first blank line is considered to be a "summary". Generally, we try to keep the summary short. It’s non-blocking.

ffi/src/winscard/scard.rs
```diff
@@ -40,34 +118,60 @@ pub extern "system" fn SCardReconnect(
...
-pub extern "system" fn SCardEndTransaction(_handle: ScardHandle, _dw_disposition: u32) -> ScardStatus {
-    todo!()
+pub unsafe extern "system" fn SCardEndTransaction(handle: ScardHandle, _dw_disposition: u32) -> ScardStatus {
+    check_handle!(handle);
+    let scard = try_execute!(scard_handle_to_winscard(handle));
+
+    try_execute!(scard.end_transaction());
+
+    ErrorKind::Success.into()
 }
```
CBenoit, line 164: suggestion: Ideally, I would like unsafe blocks even inside "unsafe function": https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unsafe-op-in-unsafe-fn
This will be warn-by-default or even deny-by-default in future editions: https://github.com/rust-lang/rust/issues/71668

note: The function should be unsafe only if the caller must unhold invariants himself. In this case, if there is an invariant on `ScardHandle`, then the function being marked unsafe is good. A typical example is when the caller must ensure a pointer is non-null. If `ScardHandle` is always valid by definition (e.g.: the user can’t craft an invalid `ScardHandle` and we always return a valid one), then the function is not unsafe to call per-se.

ffi/src/winscard/buff_alloc.rs
CBenoit, line 1: nitpick: The common abbreviation for "buffer" is "buf"

CBenoit (approved):
Thank you! Well documented again. :+1: 
I left a few comments that you can address and merge yourself. (Since we are merging into a dev branch, I think you should have the permissions.)

note: The commit type as per conventional commit specification for refactoring is `refactor` instead `refac`. If you want to stick to the convention.

crates/ffi-types/src/winscard/functions.rs
TheBestTvarynka: > Does it change the semantic

yes. When I remove the unsafe `keyword` in such places, then it won't compile.

image

ffi/src/winscard/scard_handle.rs
TheBestTvarynka: fixed

ffi/src/winscard/buff_alloc.rs
TheBestTvarynka: renamed

ffi/src/winscard/scard.rs
TheBestTvarynka: Good point! I thought about it for a while and made the following changes:

1. I made `scard_handle_to_winscard`, `scard_context_to_winscard_context`, `copy_buff`, and other similar functions also `unsafe`. They all work with raw pointers provided by the user and we can't guarantee any safe behaviour.
2. I added the `#[deny(unsafe_op_in_unsafe_fn)]` lint but for the `winscard` module only. I think we can refactor the `sspi` module as well, but it's better to do it in a separate PR.

TheBestTvarynka:
I answered the comments but didn't resolve them and left them to you. Maybe you will have any other questions

ffi/src/winscard/scard.rs
CBenoit: Sounds good to me! Thank you!!

PR #206, opened 2024-02-05
Smart card credentials handling

TheBestTvarynka:
Hi,
In this PR I implemented the smart card credentials handling on the SSPI side. How it works: if the provided username is `NT_ENTERPRISE` and all needed environment variables with smart card credentials are present, then the SSPI will try to authenticate using an emulated smart card.
Also, I made a small refactoring in the _./ffi/src/sspi_ module.

All needed env vars are listed in this module of the `winscard` crate: https://github.com/Devolutions/sspi-rs/blob/winscard-rs-impl/crates/winscard/src/env.rs.

This is the last PR with a new implemented logic in the scope of this winscard implementation and integration.

Here is a small demo:

https://github.com/Devolutions/sspi-rs/assets/43034350/0ce67179-6939-4c27-b850-fa2bf4700f06

It took a username and PIN code (password) from the UI and the rest parameters (container name, reader name, certificate, private file, and so on) from environment variables.

cc @awakecoding


crates/winscard/src/scard.rs
```diff
@@ -342,7 +347,7 @@ impl SmartCard<'_> {
         // NIST.SP.800-73-4, Part 1, Table 5
         const RSA_ALGORITHM: u8 = 0x07;
         // NIST.SP.800-73-4, Part 1, Table 4b
-        const PIV_AUTHENTICATION_KEY: u8 = 0x9A;
+        const PIV_AUTHENTICATION_KEY: u8 = 0x9C;
```
TheBestTvarynka, line 350: According to the specification, `9c` corresponds to the _Digital Signature Key_ and `9a` to the `PIV Authentication Key`.
I decided to change this value to `9c`

src/kerberos/mod.rs
```diff
@@ -873,9 +885,9 @@ impl<'a> Kerberos {
                 // the original flag is
                 // GSS_C_MUTUAL_FLAG | GSS_C_REPLAY_FLAG | GSS_C_SEQUENCE_FLAG | GSS_C_CONF_FLAG | GSS_C_INTEG_FLAG
                 // we want to be able to turn of sign and seal, so we leave confidentiality and integrity flags out
-                let flags = builder.context_requirements.into();
-                let mut checksum_value = ChecksumValues::default();
-                checksum_value.set_flags(flags);
+                // let flags = builder.context_requirements.into();
+                let checksum_value = ChecksumValues::default();
+                // checksum_value.set_flags(flags);
```
TheBestTvarynka, line 890: We have a problem here. During the dev testing, I had a problem with this code. When I uncomment these `flags`, then the CredSSP server returns an `IvalidToken` error. I'm not sure why, but when commented on it, then authentication succeeded.

src/smartcard.rs
```diff
@@ -172,109 +171,9 @@ impl SmartCard {
                 Ok(signature)
             }
             SmartCardApi::PivSmartCard(ref mut scard) => {
-                // select the PIV app
```
TheBestTvarynka, line 175: I decided to remove all APDU command creation. We use emulated smart cards inside sspi, so there is no point in creating raw APDU. We can directly communicate with the smart card

crates/winscard/src/scard.rs
CBenoit: I think the name of the item should be changed too then? It’s still named `PIV_AUTHENTICATION_KEY`.

TheBestTvarynka: Oh, right. I forgot about it.
Fixed

ffi/src/sspi/sec_winnt_auth_identity.rs
```diff
@@ -582,6 +583,38 @@ pub unsafe fn unpack_sec_winnt_auth_identity_ex2_w_sized(
...
+        use winscard::SmartCardInfo;
+
+        use crate::utils::str_encode_utf16;
+        use crate::winscard::scard_context::{DEFAULT_CARD_NAME, MICROSOFT_DEFAULT_CSP};
+
+        match SmartCardInfo::try_from_env() {
+            Ok(smart_card_info) => {
+                // remove null
+                let new_len = password.as_ref().len() - 2;
+                password.as_mut().truncate(new_len);
```
CBenoit, line 598: nitpick: I recommend writing ["proper sentences"](https://github.com/Devolutions/IronRDP/blob/60080191aa34a47e075ee7692238f34220c6092f/STYLE.md#inline-code-comments-are-proper-sentences) to  create a more appropriate frame of mind, and the expectation is to write down more of the context you had in mind when writing this code.

For instance, in this case what I really want to know as someone non familiar with the smart card protocol is "why we need to remove the null terminator".

This comment can be applied at several places, but I leave you decide if and how much you do it.

(All that being said, the code you wrote is generally well documented, thank you again.)

src/kerberos/client/generators.rs
```diff
@@ -377,6 +377,7 @@ impl From<[u8; 24]> for ChecksumValues {
 }
 
 impl ChecksumValues {
+    #[allow(dead_code)]
     pub(crate) fn set_flags(&mut self, flags: GssFlags) {
         let flag_bits = flags.bits();
         let flag_bytes = flag_bits.to_le_bytes();
```
CBenoit, line 383: nitpick: In general I prefer removing the dead code, but maybe this should be made public? I’m totally fine with removing this code too.

src/kerberos/mod.rs
```diff
@@ -769,8 +769,16 @@ impl<'a> Kerberos {
                     #[cfg(feature = "scard")]
                     CredentialsBuffers::SmartCard(smart_card) => {
                         let pin = utf16_bytes_to_utf8_string(smart_card.pin.as_ref()).into_bytes();
+                        warn!(?pin, pin2 = ?smart_card.pin);
```
CBenoit, line 772: question: Is it intended to be a warning? This sounds like reaching this code path is not great.

CBenoit (commented):
Looking quite good. We’ll have to look the Kerberos flag thing before merging though. I don’t know if this is okay to just comment out this code.

src/kerberos/mod.rs
TheBestTvarynka: It's a leftover. I thought I removed it...
Fixed.

src/kerberos/client/generators.rs
TheBestTvarynka: This code became unused because I commented `flags` in the Kerberos implementation. I decided to keep it because it can be needed when we find out the cause of the CredSSP error

ffi/src/sspi/sec_winnt_auth_identity.rs
TheBestTvarynka: I agree with you. Can you check the last commit? I updated some comments

src/kerberos/client/generators.rs
CBenoit: Oh okay! Can you add a FIXME / TODO comment on the same line?

ffi/src/sspi/sec_winnt_auth_identity.rs
CBenoit: It’s looking much better! For this place specifically, I would like something more like that:

> The "pin" buffer of the SmartCardIdentityBuffers structure must not be NULL-terminated because / as specified in […] (protocol spec reference? empirical observation?). 
> Since the password data returned by CredUnPackAuthenticationBufferW is a NULL-terminated wide C string, we need to remove the last two bytes.

The conversion itself is a detail, and what is important is the surroundings, the things we can’t see just by looking at the code.

(Again, I’m really being pedantic here.)

src/kerberos/mod.rs
CBenoit: For now, can you leave a FIXME at this place? We’ll have to address this before merging into master.

CBenoit (approved):
LGTM!

I’ll approve the changes as-is.
The flag problem mentioned [here](https://github.com/Devolutions/sspi-rs/pull/206#discussion_r1479032012) must be addressed before we merge into master, but it’s okay to merge this PR into the winscard dev branch when you are ready @TheBestTvarynka 
If you can add a FIXME just to not forget about it before, that would be perfect.

Thank you!

src/kerberos/mod.rs
TheBestTvarynka: I added a comment here.
Can I take this issue today? I want to merge smart card implementation into a master quickly.

ffi/src/sspi/sec_winnt_auth_identity.rs
TheBestTvarynka: > Again, I’m really being pedantic here.

I can understand you and I like it 😃.

I added more comprehensive explanations

src/kerberos/client/generators.rs
TheBestTvarynka: sure. done

PR #215, opened 2024-03-07
build: fix dns_resolver feature for older toolchains

CBenoit:


Cargo.toml
```diff
@@ -26,7 +26,7 @@ exclude = ["tools/sspi-ffi-attacker", "tools/wasm-testcompile"]
 [features]
 default = []
 network_client = ["dep:reqwest", "dep:portpicker"]
-dns_resolver = ["dep:trust-dns-resolver", "dep:tokio", "tokio/rt-multi-thread"]
+dns_resolver = ["dep:trust-dns-resolver", "dep:tokio"]
```
CBenoit, line 29: This seems to be okay on the latest toolchain, because the `tokio` dependency is enabled by the same feature, but on older toolchains it should be `tokio?/rt-multi-thread` (note the `?`).

Since the feature is always enabled, I just added it directly below at the place where the `tokio` dependency is declared.

awakecoding (approved):

PR #216, opened 2024-03-12
Fix TLS decryption error

### LINKED ISSUE #153: Fix mstsc.exe SChannel decryption failure on client
During the first few seconds of the RDP connection, mstsc.exe with an injected sspi-rs fails to decrypt traffic and terminates:

image

image

It's unclear what the problem is, but it prevents a successful mstsc.exe connection using sspi-rs, and it happens every time for me.

TheBestTvarynka:
Hi,
In this pull request, I have fixed the decryption error. This error used to look like this:

image

If you think "_This pull request does not look like a fix_", then I can understand you 😅 

Actually, to fix all the problems I was forced to change the TLS decryption algorithm a lot, fix the memory leaks, change the API, and other things. I recommend you review it commit by commit. I made comprehensive commit messages with problem explanations and descriptions of each fix. Also, I would like to merge and rebase instead of merge and squash this pull request. It would be great to save those explanations in the commit history.

The proposed changes contain some undocumented anywhere behavior. I added comments where appropriate. Feel free to ask any questions.

closes #153

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -127,32 +127,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
-name = "aws-lc-rs"
-version = "1.6.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df33e4a55b03f8780ba55041bc7be91a2a8ec8c03517b0379d2d6c96d2c30d95"
-dependencies = [
- "aws-lc-sys",
- "mirai-annotations",
- "paste",
- "zeroize",
-]
-
-[[package]]
-name = "aws-lc-sys"
-version = "0.13.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a5f269b176dc4aeb593910fa56ed6f956cde19542e496bb0bfc1ad9a6ce18815"
-dependencies = [
- "bindgen",
- "cmake",
- "dunce",
- "fs_extra",
- "libc",
- "paste",
-]
-
-[[package]]
 name = "backtrace"
 version = "0.3.69"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -184,29 +158,6 @@ name = "base64ct"
 version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8c3c1a368f70d6cf7302d78f8f7093da241fb8e8807c05cc9e51a125895a6d5b"
-
-[[package]]
-name = "bindgen"
-version = "0.69.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a00dc851838a2120612785d195287475a3ac45514741da670b735818822129a0"
-dependencies = [
- "bitflags 2.4.2",
- "cexpr",
- "clang-sys",
- "itertools",
- "lazy_static",
- "lazycell",
- "log",
- "prettyplease",
- "proc-macro2",
- "quote",
- "regex",
- "rustc-hash",
- "shlex",
- "syn",
- "which",
-]
 
 [[package]]
 name = "bit-set"
@@ -282,18 +233,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.88"
+version = "1.0.89"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "02f341c093d19155a6e41631ce5971aac4e9a868262212153124c15fa22d1cdc"
-
-[[package]]
-name = "cexpr"
-version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6fac387a98bb7c37292057cffc56d62ecb629900026402633ae9160df93a8766"
-dependencies = [
- "nom",
-]
+checksum = "a0ba8f7aaa012f30d5b2861462f6708eccd49c3c39863fe083a308035f63d723"
 
 [[package]]
 name = "cfg-if"
@@ -309,26 +251,6 @@ checksum = "773f3b9af64447d2ce9850330c473515014aa235e6a783b02db81ff39e4a3dad"
 dependencies = [
  "crypto-common",
  "inout",
-]
-
-[[package]]
-name = "clang-sys"
-version = "1.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "67523a3b4be3ce1989d607a828d036249522dd9c1c8de7f4dd2dae43a37369d1"
-dependencies = [
- "glob",
- "libc",
- "libloading",
-]
-
-[[package]]
-name = "cmake"
-version = "0.1.50"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a31c789563b815f77f4250caee12365734369f942439b7defd71e18a48197130"
-dependencies = [
- "cc",
 ]
 
 [[package]]
@@ -411,9 +333,9 @@ dependencies = [
 
 [[package]]
 name = "crypto-mac"
-version = "0.11.0"
+version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "25fab6889090c8133f3deb8f73ba3c65a7f456f66436fc012a1b1e272b1e103e"
+checksum = "b1d1a86f49236c215f271d40892d5fc950490551400b02ef360692c29815c714"
 dependencies = [
  "generic-array",
  "subtle",
@@ -513,12 +435,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "dunce"
-version = "1.0.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "56ce8c6da7551ec6c462cbaf3bfbc75131ebbfa1c944aeaa9dab51ca1c5f0c3b"
-
-[[package]]
 name = "ecdsa"
 version = "0.16.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


src/credssp/mod.rs
```diff
@@ -734,6 +749,21 @@ impl<'a> SspiContext {
...
+        Generator::new(move |mut yield_point| async move {
+            self.initialize_security_context_impl(&mut yield_point, builder).await
+        })
+        .resolve_with_default_network_client()
+    }
+
+    pub fn change_password_sync(&mut self, builder: ChangePassword) -> crate::Result<()> {
+        Generator::new(move |mut yield_point| async move { self.change_password_impl(&mut yield_point, builder).await })
+            .resolve_with_default_network_client()
+    }
```
TheBestTvarynka, line 765: We need those two functions to eliminate lifetime errors in the FFI.

src/credssp/sspi_cred_ssp/tls_connection.rs
```diff
@@ -63,21 +136,152 @@ impl TlsConnection {
...
-                    .process_new_packets()
-                    .map_err(|err| Error::new(ErrorKind::DecryptFailure, err.to_string()))?;
+                while !tls_packet.is_empty() {
+                    let _ = tls_connection.read_tls(&mut tls_packet)?;
 
-                let mut reader = tls_connection.reader();
+                    let tls_state = tls_connection
+                        .process_new_packets()
+                        .map_err(|err| Error::new(ErrorKind::DecryptFailure, err.to_string()))?;
+                    trace!(?tls_state);
```
TheBestTvarynka, line 255: nope, it's not a leftover 😃 
It's very convenient to see how many bytes the `rustls` has managed to decrypt. Also, it shows us an overall IO state.
Sometimes, it can be very useful during debugging.

src/decrypt_buffer.rs
TheBestTvarynka, line 1: In the scope of this pull request, I have fixed the memory leak during the decryption and made it fully in-place by using references. But we still have a similar error in the `EncryptMessage` function.
Fixing the `EncryptMessage` function is out of the scope of this PR and not critical (`mstsc` works well anyway). Shout I create an issue about it?

CBenoit:
> Also, I would like to merge and rebase instead of merge and squash this pull request. It would be great to save those explanations in the commit history.

I agree with this. However, it would be best if each commit could pass the CI.
As a best effort, could you run this for each commit and amend as appropriate?

```bash
$ cargo clippy --workspace -- -D warnings && cargo test --workspace && pwsh ./tools/wasm-testcompile/check.ps1 && cargo fmt --check
```

Here is how you can do this automatically for each commit:
```bash
$ git rebase -i --exec 'cargo clippy --workspace -- -D warnings && cargo test --workspace && pwsh ./tools/wasm-testcompile/check.ps1 && cargo fmt --check' origin/master
```

On failure, the rebase will stop so that you can fix the problem and amend using `git commit --amend`.

Thank you!

Cargo.lock
CBenoit, line 1: question: Is it necessary to update the Cargo.lock as part of this PR? I would rather leave that to a separate PR (in fact, we should enable dependabot) and remove "noise" from this PR. By noise I mean that if something breaks because of a dependency, I would rather have a commit dedicated to dependencies.

TheBestTvarynka:
> could you run this for each commit and amend as appropriate?

No problem. I'll do it

TheBestTvarynka:
> Is it necessary to update the Cargo.lock as part of this PR?

Nope. I just played with different `rustls` versions. I'll revert it

TheBestTvarynka:
@CBenoit I think you can review it in the meantime when I'm fixing the CI

src/credssp/sspi_cred_ssp/tls_connection.rs
```diff
@@ -66,17 +66,21 @@ impl TlsConnection {
     pub fn decrypt_tls(&mut self, mut payload: &[u8]) -> Result<Vec<u8>> {
         match self {
             TlsConnection::Rustls(tls_connection) => {
-                tls_connection.read_tls(&mut payload)?;
+                let mut plain_data = Vec::with_capacity(payload.len());
 
-                let tls_state = tls_connection
-                    .process_new_packets()
-                    .map_err(|err| Error::new(ErrorKind::DecryptFailure, err.to_string()))?;
+                while payload.len() != 0 {
```
CBenoit: suggestion: `while !payload.is_empty()`

src/credssp/sspi_cred_ssp/tls_connection.rs
```diff
@@ -66,17 +66,21 @@ impl TlsConnection {
...
+                        .process_new_packets()
+                        .map_err(|err| Error::new(ErrorKind::DecryptFailure, err.to_string()))?;
 
-                let mut plain_data = vec![0; tls_state.plaintext_bytes_to_read()];
-                let _plain_data_len = reader.read(&mut plain_data)?;
+                    let mut reader = tls_connection.reader();
+                    let mut decrypted = vec![0; tls_state.plaintext_bytes_to_read()];
+                    let _plain_data_len = reader.read(&mut decrypted)?;
 
+                    plain_data.extend_from_slice(&decrypted);
```
CBenoit: suggestion: Instead of allocating an intermediate buffer each time, maybe you could use [`read_to_end`](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#method.read_to_end).

```rust
reader.read_to_end(&mut plain_data)?;
```

It’s also less error prone that ignoring the read length returned by `read`. (I assume the current code is not broken, but typically we should check this value.)

CBenoit (commented):
nitpick: It would be better to not include periods in the commit message. It’s not useful and increases chances of getting the message wrapped (too long). Of course, it’s not a problem in the body of the commit.

src/decrypt_buffer.rs
CBenoit: Yes please!

Cargo.toml
```diff
@@ -64,7 +64,8 @@ trust-dns-resolver = { version = "0.23", optional = true }
 portpicker = { version = "0.1", optional = true }
 num-bigint-dig = "0.8"
 tracing = "0.1"
-rustls = { version = "0.21", features = ["dangerous_configuration"], optional = true }
+rustls = { path = "../deleteIt/rustls/rustls", optional = true }
+#rustls = { version = "0.22.2", optional = true }
```
CBenoit: issue: This intermediate commit is referring to a local fork. This commit should be amended to pass the CI if we want to rebase.

ffi/src/sspi/common.rs
```diff
@@ -265,6 +275,7 @@ pub unsafe extern "system" fn EncryptMessage(
     p_message: PSecBufferDesc,
     message_seq_no: u32,
 ) -> SecurityStatus {
+    info!("threadid: {:?} {:?}", std::thread::current().id(), *ph_context);
```
CBenoit: question: Do we really need INFO-level traces of the current thread ID?

src/credssp/sspi_cred_ssp/tls_connection.rs
```diff
@@ -117,20 +153,21 @@ impl TlsConnection {
...
+                //     rustls::SupportedCipherSuite::Tls12(cipher_suite) => &cipher_suite.common.bulk,
+                //     rustls::SupportedCipherSuite::Tls13(cipher_suite) => &cipher_suite.common.bulk,
+                // };
+                // let block_size = match bulk_cipher {
+                //     rustls::BulkAlgorithm::Aes128Gcm => AES_BLOCK_SIZE,
+                //     rustls::BulkAlgorithm::Aes256Gcm => AES_BLOCK_SIZE,
+                //     // ChaCha20 is a stream cipher
+                //     rustls::BulkAlgorithm::Chacha20Poly1305 => 0,
+                // };
+                let block_size = AES_BLOCK_SIZE;
```
CBenoit: issue: This commit seems to include many changes unrelated to the Mutex addition

src/builders/accept_sec_context.rs
```diff
@@ -117,15 +98,13 @@ impl<
...
         'a,
-        AuthData,
         CredsHandle,
         WithCredentialsHandle,
         ContextRequirementsSet,
         TargetDataRepresentationSet,
         OutputSet,
     > {
         AcceptSecurityContext {
-            inner: self.inner,
```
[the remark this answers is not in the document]
CBenoit: This may be because the lifetime bounds are not properly specified.

CBenoit:
note: I see that [this commit](https://github.com/Devolutions/sspi-rs/pull/216/commits/d57b886d845fb8c7e7d2ade670f7d943c0e8c4cb) is including two different changes and both are specified using a conventional commit message, but both are included in the same commit:

image

FYI, this is not following the specification. This should be two different commits. This is defeating the purpose of being machine readable. The body may contain an extended description of the change of the commit, and optional footers with additional machine-readable metadata, but not another commit message.

src/credssp/mod.rs
```diff
@@ -502,18 +513,22 @@ impl<C: CredentialsProxy<AuthenticationData = AuthIdentity>> CredSspServer<C> {
...
-                        .with_output(&mut output_token)
-                        .execute(),
+                    AcceptSecurityContext::<
+                        '_,
+                        _,
+                        WithoutCredentialsHandle,
+                        WithoutContextRequirements,
+                        WithoutTargetDataRepresentation,
+                        WithoutOutput,
+                    >::new()
```
CBenoit, line 525: thought: Not very ergonomic. I would expect that we don’t need to specify all the generic parameters.

CBenoit:
We can try to proceed with the rebase approach this time, but I think that it would be generally simpler to just open several PRs to avoid merging intermediate commits that do not pass the CI. Feel free to open many small PRs at any time. The smaller, the quicker I’ll review and merge it.

src/credssp/sspi_cred_ssp/tls_connection.rs
TheBestTvarynka: done: https://github.com/Devolutions/sspi-rs/issues/217 

I'll link to this issue example of the solution for the decryption problem once the corresponding PR is merged.

src/decrypt_buffer.rs
TheBestTvarynka: done: https://github.com/Devolutions/sspi-rs/issues/217 

I'll link to this issue example of the solution for the decryption problem once the corresponding PR is merged.

PR #218, opened 2024-03-13
fix(ffi): use Mutex for the sspi methods syncronization

TheBestTvarynka:
Hi,
In this pull request, I've improved the FFI by synchronization its methods using `Mutex`.

According to [MSDN](https://learn.microsoft.com/en-us/windows/win32/secauthn/decryptmessage--general): 

> [EncryptMessage (General)](https://learn.microsoft.com/en-us/windows/win32/secauthn/encryptmessage--general) and DecryptMessage (General) can be called at the same time from two different threads in a single [security support provider interface](https://learn.microsoft.com/en-us/windows/win32/secgloss/s-gly) (SSPI) context if one thread is encrypting and the other is decrypting.

During the dev testing, I confirmed that these methods are called from the different threads. But our implementation has no synchronization and we can not guarantee that it doesn't break during the connection. To fix this, I wrapped `SppiContext` into a `Mutex` and implemented all needed methods and traits for this wrapper.

I decided to merge all such intermediate pull requests into a dev (`fix/tls-decryption`) branch.


CBenoit (approved):
LGTM! Thank you!

PR #219, opened 2024-03-14
fix(ffi): memory leak during message decryption

TheBestTvarynka:
Hi,
In this pull request, I have fixed the memory leak during message decryption.

According to the [MSDN](https://learn.microsoft.com/en-us/windows/win32/secauthn/decryptmessage--schannel):

> The encrypted message is decrypted in place, overwriting the original contents of its buffer.

Our previous implementation had allocated new buffers for the decrypted data. It had caused a memory leak because the caller didn't expect new allocations. This problem is fixed by introducing a special `DecryptBuffer` type that uses only mutable references to the provided buffers without any vectors and data cloning.

Note. I suggest refraining from paying a lot of attention to the `tls_connection` module, as I have implemented changes to the decryption process in the upcoming pull request.

### DIFF
--- a/examples/client.rs
+++ b/examples/client.rs
@@ -9,8 +9,8 @@ use std::net::TcpStream;
 
 use byteorder::{LittleEndian, ReadBytesExt, WriteBytesExt};
 use sspi::{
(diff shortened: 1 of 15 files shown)


TheBestTvarynka:
Wow, CI has passed from the first try 🥳

CBenoit (approved):
LGTM! :rocket:

PR #220, opened 2024-03-14
fix(sspi): partial TLS buffers decryption

TheBestTvarynka:
Hi,
In this pull request, I have fixed the partial TLS packets decryption.
_Note._ Under the _"partial TLS packet"_ term I mean buffers that contain only some part of the TLS packet. For example, the first 130 bytes, where the full TLS packet is 240.

There can be a situation when `mstsc` can pass only some part of the TLS packet as an input buffer. Previously, we passed such buffers to the `rustls` and were waiting for the rest on the next decrypt function call. But after such n-step packet decryption, we may get a situation when the decrypted data buffer is larger than the encrypted one. It makes it impossible to properly return all decrypted data. Now, the `sspi-rs` behaves as the original `SChannel`:
In a case when it gets the partial TLS message, it'll return `SECBUFFER_MISSING` with a length equal to the needed bytes amount alongside the `SEC_E_INCOMPLETE_MESSAGE` status code.

### DIFF
--- a/ffi/src/sspi/utils.rs
+++ b/ffi/src/sspi/utils.rs
@@ -18,8 +18,8 @@ pub unsafe fn transform_credentials_handle<'a>(
     }
 }
 
-// when encoding an UTF-16 character using two code units, the 16-bit values are chosen from the UTF-16 surrogate range 0xD800–0xDFFF,
-// and thus only \0 is encoded by two consecutive null bytes
+// when encoding a UTF-16 character using two code units, the 16-bit values are chosen from
+// the UTF-16 surrogate range 0xD800–0xDFFF, and thus only \0 is encoded by two consecutive null bytes.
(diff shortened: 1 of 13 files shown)


ffi/src/sspi/utils.rs
```diff
@@ -18,8 +18,8 @@ pub unsafe fn transform_credentials_handle<'a>(
     }
 }
 
-// when encoding an UTF-16 character using two code units, the 16-bit values are chosen from the UTF-16 surrogate range 0xD800–0xDFFF,
-// and thus only \0 is encoded by two consecutive null bytes
+// when encoding a UTF-16 character using two code units, the 16-bit values are chosen from
```
CBenoit, line 21: style:
```suggestion
// When encoding a UTF-16 character using two code units, the 16-bit values are chosen from
```

CBenoit (approved):
LGTM!

PR #221, opened 2024-03-14
Fix TLS decryption

### LINKED ISSUE #153: Fix mstsc.exe SChannel decryption failure on client

TheBestTvarynka:
Hi,
In this pull request, I have fixed the decryption error. This error used to look like this:

image

More information you can read in those four pull requests description:

* https://github.com/Devolutions/sspi-rs/pull/218
* https://github.com/Devolutions/sspi-rs/pull/219
* https://github.com/Devolutions/sspi-rs/pull/220

I checked the connection in the following scenarios:

* Kerberos smartcard-based logon.
* Kerberos password-based logon.
* NTLM logon.

Docs & references:

* [DecryptMessage (Schannel)](https://learn.microsoft.com/en-us/windows/win32/secauthn/decryptmessage--schannel).
* https://stackoverflow.com/a/6832633/9123725.
* https://stackoverflow.com/a/65101172.

closes #153 
cc @awakecoding


CBenoit (approved):
Great job! Thank you!

Issue #174, opened 2023-09-22
Replace `winapi` crate with `windows`/`windows-sys`


TheBestTvarynka:
Currently, in our [`ffi`](https://github.com/Devolutions/sspi-rs/tree/master/ffi) crate we are using two crates for communication with Windows API:

* [winapi](https://docs.rs/winapi/latest/winapi/)
* [windows-sys](https://docs.rs/windows-sys/latest/windows_sys/)

But the `winapi` crate is abandoned. So we need to remove the `winapi` crate and replace its functions calls with the corresponding ones from the `windows`/`windows-sys` crate

PR #226, opened 2024-03-31
refactor: replace remaining uses of winapi with windows_sys

ekse:
This replaces the remaining uses of winapi as per https://github.com/Devolutions/sspi-rs/issues/174. 

I ran the tests with the features `scard,tsssp` enabled since some of the replaced calls are behind these features.


CBenoit (approved):
LGTM! Thank you! It was less changes than I expected, nice.

PR #229, opened 2024-04-10
Fix CredSSP credentials handling when using `ClearTextPassword` to specify the password

TheBestTvarynka:
Hi,
In this PR I've made two main changes:

* Fixed the credentials handling in the CredSSP module when using `ClearTextPassword` to specify the password. Now we can connect to the RDP server using only the `.rdp` file without any credentials prompts.
* Improved the error handling in the Kerberos implementation.

### DIFF
--- a/src/kerberos/mod.rs
+++ b/src/kerberos/mod.rs
@@ -944,7 +944,11 @@ impl<'a> Kerberos {
                     .ok_or_else(|| Error::new(ErrorKind::InvalidToken, "Input buffers must be specified"))?;
                 let input_token = SecurityBuffer::find_buffer(input, SecurityBufferType::Token)?;
 
-                let neg_token_targ: NegTokenTarg1 = picky_asn1_der::from_bytes(&input_token.buffer)?;
+                let neg_token_targ = {
+                    let mut d = picky_asn1_der::Deserializer::new_from_bytes(&input_token.buffer);
+                    let neg_token_targ: NegTokenTarg1 = KrbResult::deserialize(&mut d)??;
+                    neg_token_targ
+                };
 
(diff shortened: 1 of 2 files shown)


src/kerberos/mod.rs
```diff
@@ -944,7 +944,11 @@ impl<'a> Kerberos {
                     .ok_or_else(|| Error::new(ErrorKind::InvalidToken, "Input buffers must be specified"))?;
                 let input_token = SecurityBuffer::find_buffer(input, SecurityBufferType::Token)?;
 
-                let neg_token_targ: NegTokenTarg1 = picky_asn1_der::from_bytes(&input_token.buffer)?;
+                let neg_token_targ = {
+                    let mut d = picky_asn1_der::Deserializer::new_from_bytes(&input_token.buffer);
+                    let neg_token_targ: NegTokenTarg1 = KrbResult::deserialize(&mut d)??;
+                    neg_token_targ
+                };
```
CBenoit, line 951: suggestion: This seems identical to this
```rust
let neg_token_targ: NegTokenTarg1 = picky_asn1_der::from_bytes::<KrbResult>(&input_token.buffer)??;
```
or this
```rust
let krb_result: KrbResult = picky_asn1_der::from_bytes::(&input_token.buffer)?;
let neg_token_targ: NegTokenTarg1 = krb_result?;
```

CBenoit (approved):
LGTM!

src/kerberos/mod.rs
TheBestTvarynka: It does not work:

image

It's important to call [the custom `Rusult` deserialization](https://github.com/Devolutions/picky-rs/blob/master/picky-krb/src/data_types.rs#L384-L419) instead of the default one which is used for enums

PR #235, opened 2024-04-15
fix:remove left over println

irvingoujAtDevolution:
I am really sorry!
I forget to remove the println! while debugging, removed the println!, I'll create another PR so the linter won't ignore it.


irvingoujAtDevolution:
#236 will disallow println! from merging.

pauldumais (approved):

thenextman:
@irvingoujAtDevolution @pauldumais Do we need  a new nuget package with this? Should I yank the one published earlier?

irvingoujAtDevolution:
> @irvingoujAtDevolution @pauldumais Do we need a new nuget package with this? Should I yank the one published earlier?

is it possible to republish or unpublish the same package? I think that would be a better option.

PR #232, opened 2024-04-12
fix: temporary fix for decrypt message

irvingoujAtDevolution:
To achieve non-allocation in decrypt message and maintain the behavior with Windows SSPI provider, I here by propose this temporary solution. 

## Situation
In our LDAP stack , we have our decrypt code that passes the SecurityBuffer of stream type and SecurityBuffer of data type with 0 buffer into the Decrypt function. The code is shown below
``` c#
                  Span<Helpers.SecBuffer> buffers = stackalloc Helpers.SecBuffer[2];
                    buffers[0].BufferType = (uint)SecBufferType.SECBUFFER_STREAM;
                    buffers[0].cbBuffer = (uint)data.Length;
                    buffers[0].pvBuffer = (IntPtr)dataPtr;

                    buffers[1].BufferType = (uint)SecBufferType.SECBUFFER_DATA;
                    buffers[1].cbBuffer = 0;
                    buffers[1].pvBuffer = IntPtr.Zero;
                    Console.WriteLine("=================BEFORE=======================");
                    Console.WriteLine("buffer[0] type = " + buffers[0].BufferType);
                    Console.WriteLine("buffer[0] cbBuffer = " + buffers[0].cbBuffer);
                    Console.WriteLine("buffer[0] pvBuffer = " + buffers[0].pvBuffer);
                    PrintBuffer(buffers[0].pvBuffer, buffers[0].cbBuffer);

                    Console.WriteLine("buffer[1] type = " + buffers[1].BufferType);
                    Console.WriteLine("buffer[1] cbBuffer = " + buffers[1].cbBuffer);
                    Console.WriteLine("buffer[1] pvBuffer = " + buffers[1].pvBuffer);
                    PrintBuffer(buffers[1].pvBuffer, buffers[1].cbBuffer);
                    SSPI.DecryptMessage(_context, buffers, NextRecvSeqNo());
                    byte[] unwrapped = new byte[buffers[1].cbBuffer];
                    Console.WriteLine("=================AFTER=======================");
                    Console.WriteLine("buffer[0] type = " + buffers[0].BufferType);
                    Console.WriteLine("buffer[0] cbBuffer = " + buffers[0].cbBuffer);
                    Console.WriteLine("buffer[0] pvBuffer = " + buffers[0].pvBuffer);
                    PrintBuffer(buffers[0].pvBuffer, buffers[0].cbBuffer);

                    Console.WriteLine("buffer[1] type = " + buffers[1].BufferType);
[...]

PR #238, opened 2024-04-16
Improve `SECBUFFER_STREAM` buffer handling

TheBestTvarynka:
Hi,
In this pull request, I fixed the `SECBUFFER_STREAM` buffer handling.

This is how it works: 

I've explored the `SECBUFFER_STREAM` buffer usage and discovered that it is only used to pass the data to the decrypt function. Moreover, as I understand, it never changes its size and pointer location.

The sspi takes the inner buffer from the `Stream` decrypt buffer, writes data into it, and then assigns it to the `Data` decrypt buffer. Thus, after the decryption, the `Stream` decrypt buffer will be empty, and `Data` will contain decrypted data. But during the data copying on the FFI layer, we skip the `SECBUFFER_STREAM` buffer. So, we'll have resulting buffers the same as in the Windows SSPI.

Also, I added a test in the `ffi` crate that calls the `DecryptMessage` functions and expects the same behavior as for the original Windows SSPI. 

Related discussion: #232

### DIFF
--- a/src/kerberos/mod.rs
+++ b/src/kerberos/mod.rs
@@ -1007,68 +1007,68 @@ impl SspiEx for Kerberos {
 }
 
 #[cfg(any(feature = "test_data", test))]
-pub fn test_client() -> Kerberos {
+pub mod test_data {
     use picky_krb::constants::key_usages::{ACCEPTOR_SEAL, INITIATOR_SEAL};
+    use picky_krb::crypto::CipherSuite;
 
-    Kerberos {
-        state: KerberosState::Final,
-        config: KerberosConfig {
+    use super::{EncryptionParams, KerberosConfig, KerberosState};
+    use crate::Kerberos;
+
+    const SESSION_KEY: &[u8] = &[
+        21, 56, 207, 133, 152, 47, 177, 117, 223, 235, 169, 237, 173, 202, 11, 254, 142, 185, 237, 5, 97, 79, 112, 46,
+        73, 182, 117, 0, 35, 91, 24, 66,
+    ];
+    const SUB_SESSION_KEY: &[u8] = &[
+        146, 61, 191, 46, 26, 68, 247, 94, 124, 95, 1, 190, 15, 185, 245, 64, 18, 203, 212, 49, 43, 222, 254, 217, 85,
+        222, 7, 92, 254, 153, 105, 144,
+    ];
+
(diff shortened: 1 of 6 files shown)


src/kerberos/mod.rs
```diff
@@ -1006,18 +1006,77 @@ impl SspiEx for Kerberos {
     }
 }
 
-#[cfg(test)]
-mod tests {
+#[cfg(any(feature = "test_data", test))]
+pub fn test_client() -> Kerberos {
```
pacmancoder, line 1010: nit: Maybe it would be better to place test_client and test_server inside a module instead?:
```
#[cfg(any(feature = "test_data", test))]
pub mod test_data {
    pub fn fake_client() -> Kerberos { ... }
    pub fn fake_server() -> Kerberos { ... }
}
```

src/utils.rs
```diff
@@ -241,76 +241,55 @@ pub fn get_encryption_key(enc_params: &EncryptionParams) -> Result<&[u8]> {
...
         }
 
-        let mut inner_stream_buffer = stream_buffer.take_data();
+        let data_buffer = DecryptBuffer::find_buffer_mut(buffers, SecurityBufferType::Data)?;
 
-        stream_buffer_inner = Some(DataBuffer {
-            data: inner_stream_buffer.as_mut_ptr(),
-            size: decrypted.len(),
-        });
+        let data = &mut stream_buffer[stream_buffer_len - decrypted_len..];
```
pacmancoder, line 269: I believe this subtraction could produce an underflow. Let's use `.checked_sub()` here instead or add `INVARIANTS` comment describing why underflow is not possible (see [IronRDP style docs](https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#invariants) as an example)

TheBestTvarynka: I check the lengths in the `if` block above. We throw an error if the decrypted data is bigger than the stream buffer

src/kerberos/mod.rs
TheBestTvarynka: agreed. done :heavy_check_mark:

src/utils.rs
pacmancoder: Oh, sorry, you are right! 👍

pacmancoder (approved):
LGTM!

PR #244, opened 2024-05-27
CredSSP: fix SNI in TLS client hello packet

TheBestTvarynka:
Hi,
In this PR, I've fixed the SNI in the first TLS packet. Previously, we just set `stub_string`.
image
Now the server name is correct!

cc @awakecoding


pacmancoder (approved):
High-quality code as always, looks good to me! 👍

PR #210, opened 2024-02-14
Initial winscard implementation and integration

TheBestTvarynka:
Hi,
In this pull request, I've implemented the smart card emulation in the form of the winscard crate and integrated it in the `ffi` and `sspi`. So, now we can use an emulated smart card for the RDP authorization without additional drivers, TMP smart cards, etc.

I've chosen the NIST.SP.800-73-4 as a specification to follow. So, our emulated smart is a PIV-compatible smart card.

More information you can read in those four pull requests description:

* https://github.com/Devolutions/sspi-rs/pull/202
* https://github.com/Devolutions/sspi-rs/pull/201
* https://github.com/Devolutions/sspi-rs/pull/203
* https://github.com/Devolutions/sspi-rs/pull/204
* https://github.com/Devolutions/sspi-rs/pull/205
* https://github.com/Devolutions/sspi-rs/pull/206
* https://github.com/Devolutions/sspi-rs/pull/209

Known bugs:

* https://github.com/Devolutions/sspi-rs/issues/153

This decryption error happens during the smart card logon too. Initial debugging and research were unsuccessful. Its fix will be provided in a separate pull request.

Docs & references:

* [`winscard.h`](https://learn.microsoft.com/en-us/windows/win32/api/winscard/)
* [Teleport's rdpclient](https://github.com/gravitational/teleport/tree/46c1cab7c1b59317722f0946e643bdb387f8964d/lib/srv/desktop/rdp/rdpclient)
* [Interfaces for Personal Identity Verification – Part 1: PIV Card Application Namespace, Data Model and Representation](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-73-4.pdf)

cc @awakecoding

PR #246, opened 2024-05-30
PCSC-lite bindings

TheBestTvarynka:
Hi,
I've implemented Rust bindings to the [`pcsc-lite`](https://pcsclite.apdu.fr/) library in this pull request.

This pull request is the first in a series of pull requests for system smart card support. To support system-provided smart cards, we need to somehow access them from `sspi-rs`. On Windows, we will use native WinSCard API. We plan to use the `pcsc-lite` library on Linux and macOS; to do so, we need these bindings.

_**Why can't we just use the [`pcsc-sys`](https://crates.io/crates/pcsc-sys) crate?**_

The main reason is that it links with the PCSC framework on macOS which was deprecated for years. Now Apple encourages developers to use their _CryptoTokenKit_.
I observed the `pcsc-lite` API and decided that it's easier to write our bindings than fork, patch, and maintain the `pcsc-sys` crate. I've reused some types from our `ffi-types` crate. It made my work easier.


crates/pcsc-lite-rs/build.rs
```diff
@@ -0,0 +1,18 @@
...
+        println!(
+            "cargo:rustc-link-lib={}",
+            env::var("PCSC_LIB_NAME").unwrap_or_else(|_| "pcsclite".to_string())
+        );
+    } else {
+        Config::new()
+            .atleast_version("1")
+            .probe("libpcsclite")
+            .expect("Could not find a PCSC library");
+    }
```
pacmancoder, line 17: To clarify -- will this library be statically linked to the resulting executable?

crates/pcsc-lite-rs/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "pcsc-lite-rs"
```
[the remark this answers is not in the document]
pacmancoder: Currently `pcsc-lite-rs` is only providing low-level bindings to the native `pcsc-lite` library, and therefore it feels right to use `*-sys` suffix for such crates, even if a high-level counterpart is not planned.

Maybe @CBenoit or @awakecoding will have some thoughts about that?

crates/pcsc-lite-rs/build.rs
TheBestTvarynka: Nope, it'll be dynamically linked.
The `pcsc-lite` API is stable and likely will not be changed. So, I decided to link it dynamically

crates/pcsc-lite-rs/Cargo.toml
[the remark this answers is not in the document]
TheBestTvarynka: :thinking: 
I think we can name it `pcsc-lite-sys`. What do you think?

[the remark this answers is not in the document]
pacmancoder: Sounds good 👍

crates/pcsc-lite-rs/build.rs
pacmancoder: Hm... I have a concern that when we start to use this in IronRDP or other products, 90% of the time this library will be absent on the system, and the application will simply fail to start because of that, even if the user does not need smart card auth support at all.

Maybe it is better to use `dlopen`/`dlsym` machinery for smart card auth to load `pcsc-lite` on runtime? Not only this will make `pcsc-lite` presence on the machine optional, but this will also simplify the compilation process of sspi-rs (no need for installing/compiling `pcsc-lite` on dev machine)

Alternatively, we could just statically link the library.

cc @CBenoit @awakecoding

TheBestTvarynka: > I wonder If we actually had any problems I described above 🤔

we used the `pcsc` crate only on Windows. So, the WinSCard API was always with us on Windows 😅

awakecoding: > Maybe it is better to use dlopen/dlsym machinery for smart card auth to load pcsc-lite on runtime? Not only this will make pcsc-lite presence on the machine optional, but this will also simplify the compilation process of sspi-rs (no need for installing/compiling pcsc-lite on dev machine)

static linking is not really an option for pcsc-lite, and yes, adding a runtime dependency that is likely to be missing is a concern, especially on Linux. macOS should not be much of an issue issue though, even if I'd prefer having runtime loading of the DLL. Maybe we can begin by enabling the feature at build time only for macOS, and then rework this to find and load the pcsc-lite library at runtime in a second time before trying to use it on Linux?

TheBestTvarynka: > even if I'd prefer having runtime loading of the DLL

got it. I'll update the PR tomorrow.
I already implemented a similar behavior during the dev testing. Thus it's not a big problem to add it now

TheBestTvarynka:
@pacmancoder @awakecoding 

I rewrote the bindings. Now it loads the library in the runtime. I decided to move the code into the `ffi/src/winscard/pcsc-lite`. I thing we don't need a separate crate for just type definitions and one load function.

Also, I refactored the `sspi` and now it can be compiled on macOS/Linux with the `scard` feature turned on.

cc @CBenoit

crates/pcsc-lite-rs/build.rs
TheBestTvarynka: > I'll update the PR tomorrow

done

pacmancoder (approved):
Looks good to me now! 🥇

TheBestTvarynka:
@pacmancoder I think it's better to _squash and merge_ instead of _merge_ for such intermediate PRs (even if we merge into the dev branch). It'll allow us to have one good-looking final PR which consists of all these intermediate PRs (for example, #210 )

pacmancoder:
@TheBestTvarynka I always squash & merge when merging to master, I made a simple merge because I thought it would be simpler for you to manage dev branch bisect changes, etc., you could squash them by yourself in the dev branch if this were an intention. I'll squash & merge to Apriorit dev branches from now on

PR #249, opened 2024-06-04
System-provided smart card implementation

TheBestTvarynka:
Hi,
In this pull request, I've implemented the system-provided smart card support. Now we can call system API to access system-provided smart cards.
I've also implemented dynamic `winscard.dll` loading. We can't always use the `winscard-sys` crate, because if the `winscard.dll` is hooked (in the case of [MsRdpEx](https://github.com/Devolutions/MsRdpEx)), then the `winscard-sys` will just redirect calls to our library and we'll get a stack overflow.

### DIFF
--- a/ffi/src/winscard/system_scard/card.rs
+++ b/ffi/src/winscard/system_scard/card.rs
@@ -3,9 +3,9 @@ use std::mem::size_of;
 use std::ptr::null_mut;
 use std::slice::from_raw_parts;
 
+use ffi_types::winscard::functions::SCardApiFunctionTable;
 use ffi_types::winscard::{ScardContext, ScardHandle, ScardIoRequest};
 use num_traits::ToPrimitive;
-use ffi_types::winscard::functions::SCardApiFunctionTable;
 use winscard::winscard::{
     AttributeId, ControlCode, IoRequest, Protocol, ReaderAction, ShareMode, Status, TransmitOutData, WinScard,
 };
@@ -22,7 +22,11 @@ pub struct SystemScard {
 
 impl SystemScard {
     pub fn new(h_card: ScardHandle, h_card_context: ScardContext) -> Self {
-        Self { h_card, h_card_context, api: init_scard_api_table(), }
+        Self {
+            h_card,
+            h_card_context,
+            api: init_scard_api_table(),
+        }
     }
 }
 
@@ -41,9 +45,7 @@ impl Drop for SystemScard {
             }
             #[cfg(target_os = "windows")]
             {
-                if let Err(err) =
-                    try_execute!(unsafe { (self.api.SCardDisconnect)(self.h_card, 0) })
-                {
+                if let Err(err) = try_execute!(unsafe { (self.api.SCardDisconnect)(self.h_card, 0) }) {
                     error!(?err, "Cannot disconnect the card");
                 }
             }
@@ -115,12 +117,7 @@ impl WinScard for SystemScard {
             }
             #[cfg(target_os = "windows")]
             {
-                try_execute!(unsafe {
-                    (self.api.SCardFreeMemory)(
-                        self.h_card_context,
-                        reader_name as *const _,
-                    )
-                })?;
+                try_execute!(unsafe { (self.api.SCardFreeMemory)(self.h_card_context, reader_name as *const _,) })?;
             }
 
             return Err(Error::new(
@@ -135,9 +132,7 @@ impl WinScard for SystemScard {
         }
         #[cfg(target_os = "windows")]
         {
-            try_execute!(unsafe {
-                (self.api.SCardFreeMemory)(self.h_card_context, reader_name as *const _)
-            })?;
+            try_execute!(unsafe { (self.api.SCardFreeMemory)(self.h_card_context, reader_name as *const _) })?;
         }
 
         let status = Status {
@@ -296,9 +291,7 @@ impl WinScard for SystemScard {
         }
         #[cfg(target_os = "windows")]
         {
-            try_execute!(unsafe {
-                (self.api.SCardEndTransaction)(self.h_card, disposition.into())
-            })
+            try_execute!(unsafe { (self.api.SCardEndTransaction)(self.h_card, disposition.into()) })
         }
     }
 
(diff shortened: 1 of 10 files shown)


ffi/src/winscard/system_scard/card.rs
```diff
@@ -0,0 +1,403 @@
...
+        try_execute!(
+            unsafe { (self.api.SCardDisconnect)(self.h_card()?, disposition.into()) },
+            "SCardDisconnect failed"
+        )?;
+
+        // Mark the current card handle as disconnected.
+        self.h_card = HandleState::Disconnected;
+
+        Ok(())
+    }
```
CBenoit, line 402: praise: Thank you for properly documenting the unsafe blocks!

question: Can you make sure that clippy recognize the comment by adding [this attribute](clippy::undocumented_unsafe_blocks) on the function (no need to commit this specifically):
```rust
#[warn(clippy::undocumented_unsafe_blocks)]
```

ffi/src/winscard/system_scard/card.rs
```diff
@@ -0,0 +1,403 @@
...
+        }
+    }
+}
+
+impl Drop for SystemScard {
+    fn drop(&mut self) {
+        // The smart card handle can be explicitly disconnected before. So, there is no point
+        // in double disconnecting.
+        // Hint: It's always better to explicitly disconnect the card because the user can not pass
+        // the custom `dwDisposition` parameter in `SCardDisconnect` function.
```
CBenoit, line 85: nitpick: This comment could be used as proper documentation for the `SystemScard` struct using actual docstrings at the appropriate place. Maybe on `SystemScard` or on the `disconnect` function?

ffi/src/winscard/system_scard/card.rs
```diff
@@ -0,0 +1,403 @@
...
+        })
+    }
+
+    fn h_card(&self) -> WinScardResult<ScardHandle> {
+        if let HandleState::Connected(handle) = self.h_card {
+            Ok(handle)
+        } else {
+            Err(Error::new(
+                ErrorKind::InvalidHandle,
+                "Smart card is not connected or has been disconnected.",
```
CBenoit, line 74: style: https://github.com/Devolutions/IronRDP/blob/63743e9fcc5e3adf37e4745cc1e3ba7858e16c39/STYLE.md#formatting-of-error-messages

```suggestion
                "smart card is not connected or has been disconnected",
```

ffi/src/winscard/system_scard/card.rs
```diff
@@ -0,0 +1,403 @@
...
+enum HandleState {
+    /// The card is not connected or has been disconnected.
+    Disconnected,
+    /// The card is connected and ready to use.
+    Connected(ScardHandle),
+}
+
+pub struct SystemScard {
+    h_card: HandleState,
+    h_card_context: ScardContext,
```
CBenoit, line 35: question: What does the `h_` prefix stands for?

ffi/src/winscard/system_scard/context.rs
```diff
@@ -0,0 +1,679 @@
...
+        reader_name: &str,
+        share_mode: ShareMode,
+        protocol: Option<Protocol>,
+    ) -> WinScardResult<ScardConnectData> {
+        // SAFETY:
+        // https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new
+        // > This function will return an error if the supplied bytes contain an internal 0 byte.
+        //
+        // The Rust string slice cannot contain 0 bytes. So, it's safe to unwrap it.
+        let c_string = CString::new(reader_name).expect("Rust string slice should not contain 0 bytes");
```
CBenoit, line 83: issue: Actually, Rust strings can have interior nul-bytes, and that’s why `CString::new` is returning a `Result`.
It’s actually quite simple to make this code panic: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=02069fefe5ff543790ce894a53dfa4eb

```rust
let foo = "hello\0world";
std::ffi::CString::new(foo).unwrap();
```

CBenoit, line 83: style: `SAFETY:` comments should be used for `unsafe` blocks only. Of coure, comments explaining why the code does not panic are very welcomed too, but causing a panic is not unsafe.

ffi/src/winscard/system_scard/mod.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    Ok(res?)
+}
+
+fn parse_multi_string_owned(buf: &[u8]) -> WinScardResult<Vec<Cow<'static, str>>> {
+    Ok(parse_multi_string(buf)?
+        .into_iter()
+        .map(|r| Cow::Owned(r.to_owned()))
+        .collect())
+}
```
CBenoit, line 32: suggestion: If always returning an owned string, use `String` instead of `Cow<'a, str>`. `Cow` is used for types which may be owned or not. (`Cow<'static, str>` is a special case for types which are `'static` even if the value borrowed.)

ffi/src/winscard/system_scard/card.rs
TheBestTvarynka: Cool! I didn't know about this Clippy lint.
done

ffi/src/winscard/system_scard/card.rs
TheBestTvarynka: Good point. Probably no one will ever read the comment in the `Drop` trait implementation.
done

ffi/src/winscard/system_scard/card.rs
TheBestTvarynka: Done.
I forgot about this again :persevere:

ffi/src/winscard/system_scard/card.rs
TheBestTvarynka: `h_` stands for _handle_. I wanted to mark somehow that the value is just a handle to the corresponding card or card context. I saw a similar naming in the MSDN. For example, the [SCardConnectW](https://learn.microsoft.com/en-us/windows/win32/api/winscard/nf-winscard-scardconnectw) function.
Also, we already have a lot of things called `scard` and `scard_context`. It can confuse the user why we have the scard inside another scard, and so on.

ffi/src/winscard/system_scard/context.rs
TheBestTvarynka: :open_mouth:
You are right. I totally forgot about it.
I've replaced it with the proper error handling

ffi/src/winscard/system_scard/mod.rs
TheBestTvarynka: Yes, I agree with you. I have chosen the `Cow` instead of `String` because later the returned value will be transformed to the owned `Cow` anyway (because of the `WinSCardContext` trait API)

CBenoit (approved):
LGTM! Feel free to merge yourself.

PR #251, opened 2024-06-10
Fix null check in `save_out_buf` and `save_out_buf_wide` functions

TheBestTvarynka:
Hi,
After the last intermediate PR I did some dev testing to check if everything is sill working. I found a small mistake and fixed it in this PR.
Also, I've improved the logging a little bit.


CBenoit (approved):
LGTM! Good catch.

PR #254, opened 2024-06-22
fix: fix memory leak during encryption

### LINKED ISSUE #217: Memory leak in EncryptMessage function
According to the documentation, all encryption should be in place ([src](https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-encryptmessage)):

> The message is encrypted in place, overwriting the original contents of the structure.

But the current implementation can allocate new buffers:

https://github.com/Devolutions/sspi-rs/blob/00f0fcc3c484d8bdda2bae9957e45004768fef3d/ffi/src/sspi/common.rs#L290

and then:

https://github.com/Devolutions/sspi-rs/blob/00f0fcc3c484d8bdda2bae9957e45004768fef3d/ffi/src/sspi/sec_buffer.rs#L49

TheBestTvarynka:
Hi,
In this pull request, I've fixed the memory leak during data encryption.

My approach is the same, as I did for the decryption (#219): now we use buffers with references to the input data instead of data copying.

Don't be afraid of so many files changed. I just renamed two commonly used structures:

* `SecurityBuffer` -> `OwnedSecurityBuffer`. It uses `Vec` inside, so I've added the `Owned` prefix.
* `DecryptBuffer` -> `SecurityBuffer`. We use it for decryption and encryption, so `DecryptBuffer` is a bad name.

Moreover, I added a test for testing encryption and decryption in the FFI layer. I highly recommend running FFI tests with [`Miri`](https://github.com/rust-lang/miri).

> [!NOTE]
> Currently, we cannot run Miri because one dependency in the dependency tree has failed to compile with nightly Rust. 

I already fixed it and found some UB and memory leaks. I'll provide a separate PR with an updated `Cargo.lock` and another PR for UB + memory leak fixes.

closes #217


src/security_buffer.rs
```diff
@@ -0,0 +1,240 @@
...
+
+use crate::{Error, ErrorKind, Result, SecurityBufferType};
+
+/// A special security buffer type is used for the data decryption. Basically, it's almost the same
+/// as [OwnedSecurityBuffer] but for decryption.
+///
+/// [DecryptMessage](https://learn.microsoft.com/en-us/windows/win32/secauthn/decryptmessage--general)
+/// "The encrypted message is decrypted in place, overwriting the original contents of its buffer."
+///
+/// So, the already defined [SecurityBuffer] is not suitable for decryption because it uses [Vec] inside.
```
CBenoit, line 12: suggestion: I understand we should modify this docstring
```suggestion
/// So, the [OwnedSecurityBuffer] is not suitable for decryption because it uses [Vec] inside.
```

CBenoit:
> Moreover, I added a test for testing encryption and decryption in the FFI layer. I highly recommend running FFI tests with [Miri](https://github.com/rust-lang/miri).

Definitely. I’ll update the CI to add this once the next PR is landed.

src/security_buffer.rs
TheBestTvarynka: Oh, I thought the IDE had updated them all. I'll review them

CBenoit (approved):
LGTM! Thank you!

PR #255, opened 2024-06-25
build(deps): update Cargo.lock

TheBestTvarynka:
Hi,
In this PR, I've updated the `Cargo.lock`. I didn't update any dependencies explicitly in any `Cargo.toml` file. I just regenerated the `Cargo.lock` file.

The only purpose of this PR is to be able to compile the library with nightly Rust. One of the dependencies was failing to compile before this update:

image

### Why do we need to compile the library with nightly Rust?

We want to run FFI tests with [Miri](https://github.com/rust-lang/miri). But Miri requires a nightly toolchain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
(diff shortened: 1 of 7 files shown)


CBenoit (approved):
LGTM!

PR #256, opened 2024-06-25
Fix UB and memory leaks found by Miri

TheBestTvarynka:
Hi,
In this PR, I've fixed some UB and memory leaks I found using [Miri](https://github.com/rust-lang/miri).

### DIFF
--- a/ffi/src/sspi/common.rs
+++ b/ffi/src/sspi/common.rs
@@ -491,7 +491,9 @@ mod tests {
         };
 
         let status = unsafe { super::DecryptMessage(&mut kerberos_client_context, &mut message, 0, null_mut()) };
+        assert_eq!(status, 0);
 
+        let status = unsafe { super::DeleteSecurityContext(&mut kerberos_client_context) };
         assert_eq!(status, 0);
 
         // Check SECBUFFER_STREAM
@@ -546,7 +548,9 @@ mod tests {
         };
 
         let status = unsafe { super::EncryptMessage(&mut kerberos_server_context, 0, &mut message, 0) };
+        assert_eq!(status, 0);
 
+        let status = unsafe { super::DeleteSecurityContext(&mut kerberos_server_context) };
         assert_eq!(status, 0);
 
         let mut kerberos_client_context = kerberos_sec_handle(kerberos_client);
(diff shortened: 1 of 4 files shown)


ffi/src/sspi/common.rs
```diff
@@ -364,10 +364,13 @@ unsafe fn p_sec_buffers_to_decrypt_buffers(raw_buffers: &[SecBuffer]) -> sspi::R
...
-            buf.with_data(unsafe {
-                from_raw_parts_mut(raw_buffer.pv_buffer as *mut u8, raw_buffer.cb_buffer.try_into()?)
-            })?
+            let data = if raw_buffer.pv_buffer.is_null() || raw_buffer.cb_buffer == 0 {
+                &mut []
+            } else {
+                // SAFETY: the safety contract [raw_buffers] must be upheld by the caller.
+                unsafe { from_raw_parts_mut(raw_buffer.pv_buffer as *mut u8, raw_buffer.cb_buffer.try_into()?) }
+            };
+            buf.with_data(data)?
```
TheBestTvarynka, line 373: The caller can pass en empty buffer type with null pointer and zero length. Creating a slide from `null` and `0` is UB

ffi/src/sspi/sec_pkg_info.rs
```diff
@@ -161,7 +161,7 @@ pub unsafe extern "system" fn EnumerateSecurityPackagesA(
         let mut size = size_of::<SecPkgInfoA>() * packages.len();
 
         for package in &packages {
-            size += package.name.as_ref().as_bytes().len() + package.comment.as_bytes().len();
+            size += package.name.as_ref().as_bytes().len() + 1 /* null byte */ + package.comment.as_bytes().len() + 1 /* null byte */;
```
TheBestTvarynka, line 164: We had a memory access error before

ffi/src/sspi/sec_winnt_auth_identity.rs
```diff
@@ -718,17 +719,39 @@ pub unsafe extern "system" fn SspiEncodeStringsAsAuthIdentity(
...
-            domain: psz_domain_name,
-            domain_length: domain.len().try_into().unwrap(),
-            password: psz_packed_credentials_string,
-            password_length: password.len().try_into().unwrap(),
+            user,
+            user_length: user_length.try_into().unwrap(),
+            domain,
+            domain_length: domain_length.try_into().unwrap(),
+            password,
+            password_length: password_length.try_into().unwrap(),
```
TheBestTvarynka, line 754: we shouldn't use provided buffers (because we don't manage their lifetime) here but copy the data and then use it

CBenoit (approved):
Nice findings!

PR #259, opened 2024-07-02
Move OS-independent smart card code out of `scard` feature

probakowski:
This PR aims to move code that doesn't depend on WinAPI (emulated smart card, everything in the protocol etc) out of `scard` feature. It can be useful in different context that just Windows real smart card integration.

It also fixes couple of bugs I found along the way:
- discrepancies between string being treated as UTF-16 or UTF-8 (I tried to make it consistent with field docs)
- `PIV_DIGITAL_SIGNATURE_KEY` value, I made it consistent with comment above and error message


crates/winscard/Cargo.toml
```diff
@@ -11,6 +11,7 @@ description = "A Rust implementation of WinSCard"
 
 [features]
 std = ["tracing/std", "base64", "picky-asn1-der"]
+scard = ["dep:flate2"]
```
TheBestTvarynka, line 14: What is the point of the `scard` feature for the `winscard` crate?
I mean, if the user uses the `winscard` crate, then, obviously, the scard is performed.

crates/winscard/src/lib.rs
```diff
@@ -11,13 +11,15 @@ mod macros;
...
+#[cfg(feature = "scard")]
 mod compression;
 mod dummy_rng;
 /// Contains env variables names that represent smart card credentials.
 #[cfg(feature = "std")]
 pub mod env;
 mod piv_cert;
 mod scard;
+#[cfg(feature = "scard")]
 mod scard_context;
```
TheBestTvarynka, line 23: Why did you make the `scard_context` optional? I mean that the `winscard` crate would be useless if the user will not be able to acquire the scard context.

crates/winscard/src/scard.rs
```diff
@@ -362,7 +362,7 @@ impl SmartCard<'_> {
         // NIST.SP.800-73-4, Part 1, Table 5
         const RSA_ALGORITHM: u8 = 0x07;
         // NIST.SP.800-73-4, Part 1, Table 4b
-        const PIV_DIGITAL_SIGNATURE_KEY: u8 = 0x9C;
+        const PIV_DIGITAL_SIGNATURE_KEY: u8 = 0x9A;
```
TheBestTvarynka, line 365: The value of this constant is correct. The scard auth using `mstsc.exe` will break if we change it. I suppose the name is bad. I think we can rename it, but we can't change its value

ffi/src/winscard/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(all(target_os = "windows", feature = "scard"))]
```
TheBestTvarynka, line 1: currently, we should be able to compile this module on non-Windows OS. Do you have any problems with that?

ffi/src/winscard/pcsc_lite/functions.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(all(target_os = "windows", feature = "scard"))]
```
TheBestTvarynka, line 1: this module is supposed to be compiled on Linux and macOS. Why did you make it Windows-only?

TheBestTvarynka: > this module is supposed to be compiled on Linux and macOS

the idea is to be able to use system-provided smart cards on Linux and macOS using the pcsc-lite library

src/auth_identity.rs
```diff
@@ -260,6 +261,8 @@ mod scard_credentials {
     pub struct SmartCardIdentity {
         /// Username
         pub username: String,
+        /// Domain
+        pub domain: String,
```
TheBestTvarynka, line 265: Why do you need a domain here? usually, the domain is extracted from the FQDN username or user certificate.

TheBestTvarynka (commented):
Hi
Thank you for paying attention to this project and for your effort.
I have some questions about the code and your approach 🙃

probakowski:
@TheBestTvarynka I think I either misunderstood big part of the code or started with older version and didn't do a good job when merging. Let me close this one for now and get back to drawing board

TheBestTvarynka:
@probakowski no problem.
Also, you can ping me and ask for any information regarding the code :slightly_smiling_face:

PR #260, opened 2024-07-08
refactor: move hardware-independent code out of scard feature

probakowski:
This PR aims to move code that doesn't depend on real smart card reader out of `scard` feature.

### DIFF
--- a/src/kerberos/mod.rs
+++ b/src/kerberos/mod.rs
@@ -11,6 +11,7 @@ use std::io::Write;
 
 pub use encryption_params::EncryptionParams;
 use lazy_static::lazy_static;
+use picky::key::PrivateKey;
 use picky_asn1::restricted_string::IA5String;
 use picky_asn1::wrapper::{ExplicitContextTag0, ExplicitContextTag1, OctetStringAsn1, Optional};
 use picky_asn1_x509::oids;
@@ -22,7 +23,7 @@ use picky_krb::gss_api::{NegTokenTarg1, WrapToken};
 use picky_krb::messages::{ApReq, AsRep, KdcProxyMessage, KdcReqBody, KrbPrivMessage, TgsRep};
 use rand::rngs::OsRng;
 use rand::Rng;
-#[cfg(feature = "scard")]
+use rsa::{Pkcs1v15Sign, RsaPrivateKey};
 use sha1::{Digest, Sha1};
 use url::Url;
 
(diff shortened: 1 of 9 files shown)


probakowski:
@TheBestTvarynka this is (much leaner and more correct probably) version of https://github.com/Devolutions/sspi-rs/pull/259 if you have some time for a review

TheBestTvarynka:
>  if you have some time for a review

I'll review it today

src/kerberos/mod.rs
```diff
@@ -745,56 +736,43 @@ impl<'a> Kerberos {
...
+                                kdc_req_body: &kdc_req_body,
+                                dh_parameters: self.dh_parameters.clone().unwrap(),
+                                sign_data: Box::new(move |data_to_sign| {
+                                    let mut sha1 = Sha1::new();
+                                    sha1.update(data_to_sign);
+                                    let hash = sha1.finalize().to_vec();
+                                    let private_key = PrivateKey::from_pem_str(&private_key_pem)?;
+                                    let rsa_private_key = RsaPrivateKey::try_from(&private_key)?;
+                                    Ok(rsa_private_key.sign(Pkcs1v15Sign::new::<Sha1>(), &hash)?)
+                                }),
```
TheBestTvarynka, line 770: I'm concerned a bit with removing the _smartcard_ layer and just using the simple RSA signature generation. But I think it's not a problem for us. Because of two reasons:

* The `winscard` crate does the same.
* Anyway, we don't support system-provided smart cards in our current sspi implementation. But we will in the next releases and I'll look into this code and refactor it.

cc @awakecoding @CBenoit

TheBestTvarynka (approved):
Nice improvements :+1:

src/smartcard.rs
```diff
@@ -54,6 +54,7 @@ impl SmartCard {
         })
     }
 
+    #[allow(dead_code)]
```
CBenoit, line 57: question: Why this code is dead? Can you add a comment explaining why we should keep the code around? Thank you!

src/kerberos/mod.rs
CBenoit: Thank you for the heads-up! If you’re going to look into it afterwards, I’m fine with that :+1:

src/smartcard.rs
TheBestTvarynka: this one is also related to [my comment above](https://github.com/Devolutions/sspi-rs/pull/260#discussion_r1672888320):

> ...removing the smartcard layer...

CBenoit: Indeed! In general, we avoid keeping dead code around so if we decide to do so here, I would like a FIXME comment so we don’t forget forever about it. In this case, I believe you’ll end up re-using these functions later so it’s fine.

src/kerberos/mod.rs
probakowski: The aim was to cut the dependency on `winscard` crate and to avoid including unused libraries for system-provided smart cards when it's not needed. So yeah, the code is basically copied over from emulated part of `winscard` crate.

src/smartcard.rs
probakowski: I've added FIXME

probakowski:
@CBenoit I merged `fix-windows-build` branch into mine so for now github shows changes from both, after that PR lands it should go back to normal

CBenoit (approved):
LGTM! Thank you!
(I rebased the PR on master to fix the conflicts and retrieve the Windows build fix.)

PR #263, opened 2024-07-10
build: fix Windows build for newer rustls

CBenoit:
I don’t know why, but this PR got merged even though CI wasn’t green: https://github.com/Devolutions/sspi-rs/pull/261

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -116,6 +116,9 @@ jobs:
           Write-Host "Installed NASM: $(nasm --version)"
           Write-Host "Installed ninja: $(ninja --version)"
 
+          # This is required to avoid linker errors when targetting Windows ARM.
+          # Indeed libclang / LLVM is a link-time requirement for AWS LC (or ring).
+          # https://aws.github.io/aws-lc-rs/requirements/windows.html#libclang--llvm
           $VSINSTALLDIR = $(vswhere.exe -latest -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath)
           $VCINSTALLDIR = Join-Path $VSINSTALLDIR "VC"
           $LLVM_ROOT = Join-Path $VCINSTALLDIR "Tools\Llvm\x64"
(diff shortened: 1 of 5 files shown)


.github/workflows/build-native.yml
```diff
@@ -104,21 +104,13 @@ jobs:
...
-
-          $VSINSTALLDIR = $(vswhere.exe -latest -requires Microsoft.VisualStudio.Component.VC.Llvm.Clang -property installationPath)
-          $VCINSTALLDIR = Join-Path $VSINSTALLDIR "VC"
-          $LLVM_ROOT = Join-Path $VCINSTALLDIR "Tools\Llvm\x64"
-          echo "PATH=$Env:PATH;${LLVM_ROOT}\bin" >> $Env:GITHUB_ENV
-          @('', '[patch.crates-io]',
-           'ring = { git = "https://github.com/awakecoding/ring", branch = "v0.16.20-patched" }') | % {
-             Add-Content -Path "Cargo.toml" -Value $_
-           }
+          echo "PATH=$Env:PATH" >> $Env:GITHUB_ENV
```
CBenoit, line 113: cc @awakecoding 
Latest version of rustls supports two backends: ring and AWS LC. The latter being the new default.

src/credssp/sspi_cred_ssp/tls_connection.rs
```diff
@@ -318,17 +353,24 @@ impl TlsConnection {
...
                     // ChaCha20 is a stream cipher
-                    rustls::BulkAlgorithm::Chacha20Poly1305 => 0,
+                    Some(name) if name.contains("CHACHA20_POLY1305") => 0,
+                    _ => {
+                        return Err(Error::new(
+                            ErrorKind::UnsupportedFunction,
+                            format!("cipher suite {suite:?} not supported"),
+                        ))
+                    }
                 };
```
CBenoit, line 374: @TheBestTvarynka rustls API significantly changed in a breaking way and the `BulkAlgorithm` enum does not exist anymore. For the purpose of making this code compile again, I used this dirty trick, but I don’t think it’s a very good idea. Do you think you could look for a proper solution here?

TheBestTvarynka: > Do you think you could look for a proper solution here?

Yes, I can. Should I prioritize it over my current tasks or it can wait a little?

CBenoit: Thank you!
If you're almost done, go ahead with your current tasks, otherwise it would be good to take care of that in priority because I can't cut a new release with this

awakecoding (approved):

PR #272, opened 2024-07-25
fix(sspi): tls connection: improve cipher block size calculation

TheBestTvarynka:
Hi,
In this pull request, I've improved the cipher block size extraction. Related comment: https://github.com/Devolutions/sspi-rs/pull/263#discussion_r1671978937 

I browsed the new API and tried different methods. Conclusion: there is no good way of extracting the cipher block size. So, I decided to take supported cipher suites in SChannel (Windows 10/11) + popular ones and list all of them in one `match`.
Such a code is more explicit and easy to maintain.


CBenoit (approved):
This approach is satisfying. Thank you!

PR #274, opened 2024-07-26
Fix smart card private key handling

TheBestTvarynka:
Hi,
In this PR, I've fixed the scared pk handling. Now we convert it to UTF-8 as other credentials buffers.

We encountered this problem after this PR: https://github.com/Devolutions/sspi-rs/pull/260/files#diff-40e9fc919e63077bc0b25847ecf6746216d0438cd3808de82d713a7cc71bce56R753-R756 

cc @awakecoding


CBenoit (approved):
LGTM!

PR #275, opened 2024-07-26
test: turn sspi-ffi-attacker into tests runnable with miri

TheBestTvarynka:
Hi,
In the scope of this PR, I've refactored the `sspi-fii-attacker`:

1. Removed the `sspi-ffi-attacker`.
2. Moved scenarios to `ffi/sspi/sec_handle/tests`.
   Now we can run those scenarios with Miri which will help us catch UBs and other memory errors.
3. Fixed two UBs.

Additional context: https://github.com/Devolutions/sspi-rs/pull/261#discussion_r1671806468


ffi/src/sspi/sec_handle.rs
```diff
@@ -591,7 +595,11 @@ pub unsafe extern "system" fn InitializeSecurityContextW(
         let mut input_tokens = if p_input.is_null() {
             Vec::new()
         } else {
-            p_sec_buffers_to_security_buffers(from_raw_parts((*p_input).p_buffers, (*p_input).c_buffers as usize))
+            p_sec_buffers_to_security_buffers(if (*p_input).p_buffers.is_null() {
+                &[]
+            } else {
+                from_raw_parts((*p_input).p_buffers, (*p_input).c_buffers as usize)
+            })
```
TheBestTvarynka, line 602: slice creation from the null pointer is UB

ffi/src/sspi/sec_handle.rs
```diff
@@ -1212,3 +1220,274 @@ pub extern "system" fn QueryCredentialsAttributesExW(
...
+        // This test simulates initialize security context function call. It's better to run it using Miri
+        // https://github.com/rust-lang/miri
+        // cargo +nightly miri test
+        let mut pkg_name = "NTLM\0".encode_utf16().collect::<Vec<_>>();
+        let mut pkg_info: PSecPkgInfoW = null_mut::<SecPkgInfoW>();
+
+        let status = unsafe { QuerySecurityPackageInfoW(pkg_name.as_ptr(), &mut pkg_info) };
+        assert_eq!(status, 0);
+
+        println!("{:?}", unsafe { &*pkg_info });
```
TheBestTvarynka, line 1257: I left those `println!`s on purpose: just to simulate any memory access to the allocated memory.

ffi/src/sspi/sec_pkg_info.rs
```diff
@@ -22,8 +22,10 @@ pub struct SecPkgInfoW {
 
 pub type PSecPkgInfoW = *mut SecPkgInfoW;
 
+pub struct RawSecPkgInfoW(pub *mut SecPkgInfoW);
```
TheBestTvarynka, line 25: It was fun to investigate and fix. In short, conversions chain like this one `*mut SecPkgInfoW -> &mut SecPkgInfoW -> *mut SecPkgInfoW` can create UB 😮 
image_2024-07-25_22-31-46
When we create `&mut SecPkgInfoW`, Rust creates a SharedReadWrite tag (it relates to stacked borrows and so on). And when we convert it to the raw pointer and try to deallocate, we have UB

ffi/src/sspi/sec_handle.rs
CBenoit: Sounds good! Can you write this down as an in-code comment please?

ffi/src/sspi/sec_pkg_info.rs
CBenoit: Good job!

CBenoit, line 25: question/nitpick: Should we use the `PSecPkgInfoW` type alias? Remove it?

ffi/src/sspi/utils.rs
```diff
@@ -30,10 +30,22 @@ pub fn raw_wide_str_trim_nulls(raw_str: &mut Vec<u8>) {
...
-        )
-    })
+    // We run tests with Miri. Miri is the Rust's mid-level intermediate representation interpreter.
+    // It is unable to execute system calls. Thus, Miri cannot execute `whoami::fallible::hostname()`.
+    // So, we decided to keep hardcoded hostname value for tests.
+    #[cfg(test)]
+    {
+        Ok("test-vm".into())
+    }
+    #[cfg(not(test))]
```
CBenoit, line 40: suggestion: You could use the `miri` cfg instead, so we don’t stub this code when miri is not used
```suggestion
    #[cfg(miri)]
    {
        Ok("test-vm".into())
    }
    #[cfg(not(miri))]
```

ffi/src/sspi/sec_handle.rs
TheBestTvarynka: done

ffi/src/sspi/sec_pkg_info.rs
TheBestTvarynka: We are not allowed to implement the `From` trait for any raw pointers, because all raw pointers are foreign types

ffi/src/sspi/utils.rs
TheBestTvarynka: Wow, didn't know about it. Got it!

TheBestTvarynka: done

ffi/src/sspi/sec_pkg_info.rs
CBenoit: Sorry, I meant this:

```suggestion
pub struct RawSecPkgInfoW(pub PSecPkgInfoW);
```

I am under the impression that we don’t really use this alias.

CBenoit (approved):
Very nice! LGTM!

CBenoit:
Merging now, the remaining comment is not very important

PR #278, opened 2024-08-01
fix(sspi): make TLS backend configurable via features

CBenoit:
The aws-lc-rs backend is enabled by default.


TheBestTvarynka (approved):
LGTM!
I made two test runs on my environment:

* First one with `aws-lc-rs` crypto backend.
* Second one with `ring` crypto backend.

It connected successfully in both cases.

CBenoit:
Thank you!! @TheBestTvarynka
I fixed CI for Android, should be good to go.

awakecoding (approved):

PR #294, opened 2024-10-03
Fix Linux/macOS system smart card support

TheBestTvarynka:
Hi,
In this PR I've fixed the smart card auth for Linux and macOS. Now we can use smart card auth with FreeRDP on Linux and macOS.

## Implementation notes

Originally, we planned to link with `pcsc-lite` on Linux and macOS. But it turned out that the `pcsc-lite` doesn't work well on macOS. We have tried to overcome it in a few ways but failed. After all tries we decided to link with the `PCSC.framework`. We adjusted the bindings and managed to make it work.

## Configuration

To use the smart card auth you need to perform some configurations.

1. **Kerberos.** You need to write your own `krb.conf` or ask your sysadmin to provide you with one. It should look like this (replace domain, CA cert file path, and realm with your ones):
```conf
[libdefaults]
        default_realm = TBT.COM

[realms]
        TBT.COM = {
                admin_server = WIN-956CQOSSJTF.tbt.com
                kdc = WIN-956CQOSSJTF.tbt.com
                pkinit_anchors = FILE:/Users/tbt/Documents/ca_cert.cer
                pkinit_kdc_hostname = WIN-956CQOSSJTF.tbt.com
                pkinit_cert_match = <SAN>.*@tbt.com

                # The option below is important and required. The MIT KRB5 can reject the CA certificate without it. More info:
                # https://web.mit.edu/kerberos/krb5-latest/doc/admin/pkinit.html#configuring-the-clients
                # > A commercially issued server certificate will usually not have the standard PKINIT principal name or Extended Key Usage extensions
                pkinit_eku_checking = kpServerAuth
        }

[domain_realm]
        tbt.com = TBT.COM
        tbt = TBT

[logging]
        kdc = CONSOLE
```
2. **Smart card.** Insert the smart card into the Windows domain-joined machine and enroll the certificate or generate the certificate manually and import it on the smart card.
3. **Environment variables.** Some of them are optional, and some of them are required. See the table below:

| Env variable name | Meaning | Example | Required? |
| --- | --- | --- | --- |
| `WLOG_LEVEL` | `FreeRDP` logging level | trace | optional |
| `KRB5_TRACE` | log internal krb5 library operations. [docs](https://web.mit.edu/kerberos/krb5-latest/doc/admin/troubleshoot.html). | `/dev/stdout` | optional |
[...]

PR #301, opened 2024-10-08
fix(ffi): compilation with `scard` feature on macOS

TheBestTvarynka:
This problem was added in #294 . I don't know why CI didn't cache it. Maybe the `--scard` feature wasn't turned on during compilation on macOS/Linux

### DIFF
--- a/ffi/src/winscard/system_scard/card.rs
+++ b/ffi/src/winscard/system_scard/card.rs
@@ -187,7 +187,7 @@ impl WinScard for SystemScard {
             readers,
             #[cfg(not(target_os = "windows"))]
             state: crate::winscard::pcsc_lite::State::from_bits(state)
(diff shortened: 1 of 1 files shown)


TheBestTvarynka:
> Maybe the `--scard` feature wasn't turned on during compilation on macOS/Linux

https://github.com/Devolutions/sspi-rs/blob/dc61d98d1398b1dc5d310ade6d391c07b572720d/.github/workflows/ci.yml#L36-L43 and https://github.com/Devolutions/sspi-rs/blob/dc61d98d1398b1dc5d310ade6d391c07b572720d/.github/workflows/ci.yml#L65-L76

Let's enable this feature on Linux and macOS.

CBenoit (approved):

Issue #212, opened 2024-03-04
Null Identity on AcquireCredentialsHandle results in error


pauldumais:
If I pass a null identity to AcquireCredentialsHandle I get an error, which is different than what happens in the windows sspi.

AcquireCredentialsHandle failed (One or more of the parameters passed to the function was invalid., Win32ErrorCode -2146892963 - 0x8009035D)

If we pass an identity to the AcquireCredentialsHandle, but pass an empty string to the username then we get thsi error later in the process:

 InitializeSecurityContext failed (The token supplied to the function is invalid, Win32ErrorCode -2146893048 - 0x80090308)

irvingoujAtDevolution:
I could not reproduce this error on Windows, this is the code I use to try reproducing, did I miss something?
```
fn main() {
    let username = "".to_string(); // username@domain
    let password = "password".to_string(); // "password

    let kerberos_config = KerberosConfig::new("kdc_url", "hostname".to_string());
    let mut negitiate = Negotiate::new(
        NegotiateConfig::from_protocol_config(Box::new(kerberos_config.clone()), "client".to_string())
    ).unwrap();
    let _ = get_cred_handle(&mut negitiate, username, password);
}

pub(crate) fn get_cred_handle<T>(
    sspi: &mut T,
    username: String,
    password: String,
) -> AcquireCredentialsHandleResult<Option<CredentialsBuffers>>
where
    T: SspiImpl<CredentialsHandle = Option<CredentialsBuffers>> + Sspi,
    <T as SspiImpl>::AuthenticationData: From<sspi::AuthIdentity>,
{
    let identity = sspi::AuthIdentity {
        username: Username::parse(&username).expect("username is not in the correct format"),
        password: password.into(),
    };
    let acq_creds_handle_result = sspi
        .acquire_credentials_handle()
        .with_credential_use(sspi::CredentialUse::Outbound)
        .with_auth_data(&dbg!(identity).into())
        .execute()
        .expect("AcquireCredentialsHandle resulted in error");

    acq_creds_handle_result
}

```

CBenoit:
Closing as discussed in https://github.com/Devolutions/sspi-rs/pull/304#discussion_r1793546562

PR #304, opened 2024-10-09
Fix memory corruption in the `AcquireCredentialsHandleA/W` functions

TheBestTvarynka:
Hi,
In this PR I've fixed the memory corruption error in the `AcquireCredentialsHandleA/W` functions. The user can safely pass the `null`/empty credentials strings.
Also, I covered such cases with tests.

### DIFF
--- a/ffi/src/sspi/sec_handle.rs
+++ b/ffi/src/sspi/sec_handle.rs
@@ -1253,11 +1253,11 @@ mod tests {
 
     extern "system" fn dummy(_: *mut c_void, _: *mut c_void, _: u32, _: *mut *mut c_void, _: *mut i32) {}
 
+    /// This test simulates initialize security context function call. It's better to run it using Miri
+    /// https://github.com/rust-lang/miri
+    /// cargo +nightly miri test
     #[test]
     fn initialize_security_context_w() {
-        // This test simulates initialize security context function call. It's better to run it using Miri
-        // https://github.com/rust-lang/miri
-        // cargo +nightly miri test
         let pkg_name = "NTLM\0".encode_utf16().collect::<Vec<_>>();
         let mut pkg_info: PSecPkgInfoW = null_mut::<SecPkgInfoW>();
 
@@ -1379,11 +1379,11 @@ mod tests {
         assert_eq!(status, 0);
     }
 
+    /// This test simulates initialize security context function call. It's better to run it using Miri
+    /// https://github.com/rust-lang/miri
+    /// cargo +nightly miri test
     #[test]
     fn initialize_security_context_a() {
-        // This test simulates initialize security context function call. It's better to run it using Miri
-        // https://github.com/rust-lang/miri
-        // cargo +nightly miri test
         let pkg_name = "NTLM\0";
         let mut pkg_info: PSecPkgInfoA = null_mut::<SecPkgInfoA>();
 
@@ -1505,11 +1505,11 @@ mod tests {
         assert_eq!(status, 0);
     }
 
+    /// This test simulates initialize security context function call. It's better to run it using Miri
+    /// https://github.com/rust-lang/miri
+    /// cargo +nightly miri test
     #[test]
     fn acquire_credentials_handle_w_null_credentials() {
-        // This test simulates initialize security context function call. It's better to run it using Miri
-        // https://github.com/rust-lang/miri
-        // cargo +nightly miri test
         let pkg_name = "NTLM\0".encode_utf16().collect::<Vec<_>>();
 
         let user = "user".encode_utf16().collect::<Vec<_>>();
@@ -1572,11 +1572,11 @@ mod tests {
         }
     }
 
+    /// This test simulates initialize security context function call. It's better to run it using Miri
+    /// https://github.com/rust-lang/miri
+    /// cargo +nightly miri test
     #[test]
     fn acquire_credentials_handle_w_empty_credentials() {
-        // This test simulates initialize security context function call. It's better to run it using Miri
-        // https://github.com/rust-lang/miri
-        // cargo +nightly miri test
         let pkg_name = "NTLM\0".encode_utf16().collect::<Vec<_>>();
 
         let user = "".encode_utf16().collect::<Vec<_>>();
@@ -1617,11 +1617,11 @@ mod tests {
         assert_eq!(status, 0);
     }
 
+    /// This test simulates initialize security context function call. It's better to run it using Miri
+    /// https://github.com/rust-lang/miri
+    /// cargo +nightly miri test
     #[test]
     fn acquire_credentials_handle_a_null_credentials() {
-        // This test simulates initialize security context function call. It's better to run it using Miri
-        // https://github.com/rust-lang/miri
-        // cargo +nightly miri test
         let pkg_name = "NTLM\0";
 
         let user = "user";
@@ -1684,11 +1684,11 @@ mod tests {
         }
     }
 
+    /// This test simulates initialize security context function call. It's better to run it using Miri
+    /// https://github.com/rust-lang/miri
+    /// cargo +nightly miri test
     #[test]
     fn acquire_credentials_handle_a_empty_credentials() {
-        // This test simulates initialize security context function call. It's better to run it using Miri
-        // https://github.com/rust-lang/miri
-        // cargo +nightly miri test
         let pkg_name = "NTLM\0";
 
         let user = "";
--- a/ffi/src/sspi/sec_handle.rs
+++ b/ffi/src/sspi/sec_handle.rs
@@ -312,6 +312,10 @@ fn verify_security_package(package_name: &str) -> Result<()> {
     }
 }
 
+/// AcquireCredentialsHandleA function acquires a handle to preexisting credentials of a security principal.
+///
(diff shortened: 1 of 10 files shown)


ffi/src/sspi/sec_handle.rs
```diff
@@ -312,6 +312,10 @@ fn verify_security_package(package_name: &str) -> Result<()> {
     }
 }
 
+/// AcquireCredentialsHandleA function acquires a handle to preexisting credentials of a security principal.
+///
+/// NOTE: Although in the original Windows SSPI, `p_auth_data` parameter can be NULL, in our implementation it must be non-NULL.
+/// That's because we cannot get the default credentials handle for security package, like Windows can.
```
TheBestTvarynka, line 318: https://learn.microsoft.com/en-us/windows/win32/secauthn/acquirecredentialshandle--general 

> pAuthData [in]
> ...This parameter can be NULL, which indicates that the default credentials for that [security package](https://learn.microsoft.com/en-us/windows/win32/secgloss/s-gly) must be used...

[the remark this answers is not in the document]
CBenoit: If there is no way to get the default credentials, should we close #212?

ffi/src/sspi/sec_handle.rs
```diff
@@ -1494,4 +1502,228 @@ mod tests {
         let status = unsafe { DeleteSecurityContext(&mut new_sec_context) };
         assert_eq!(status, 0);
     }
+
+    #[test]
+    fn acquire_credentials_handle_w_null_credentials() {
+        // This test simulates initialize security context function call. It's better to run it using Miri
+        // https://github.com/rust-lang/miri
+        // cargo +nightly miri test
```
CBenoit, line 1510: suggestion: You can document tests with doc comments above the function
```suggestion
    /// This test simulates initialize security context function call. It's better to run it using Miri
    /// https://github.com/rust-lang/miri
    /// cargo +nightly miri test
    #[test]
    fn acquire_credentials_handle_w_null_credentials() {
```

ffi/src/utils.rs
```diff
@@ -26,8 +26,18 @@ pub unsafe fn w_str_len(s: *const u16) -> usize {
...
+///
+/// Credentials are often represented as strings. For example, username, domain, password.
+/// It is OK for Windows SSPI to accept `null` or empty credential strings. The `AcquireCredentialsHandle`
+/// function will return successful status code is we pass the `null` username value. Thus, this function
+/// will return an empty [Vec] in such a case. It is done on purpose to follow the Windows SSPI behaviour.
+pub unsafe fn credentials_str_into_bytes(raw_buffer: *const c_char, len: usize) -> Vec<u8> {
+    if !raw_buffer.is_null() {
+        unsafe { from_raw_parts(raw_buffer as *const u8, len) }.to_vec()
+    } else {
+        Vec::new()
```
CBenoit, line 39: suggestion: Can you add a safety section in the documentation? In this case, we have the same safety invariants that `from_raw_parts`: https://doc.rust-lang.org/std/vec/struct.Vec.html#safety

I also suggest to properly comment the unsafe blocks explaining how we uphold the invariants.

Examples:
- https://github.com/Devolutions/devolutions-gateway/blob/8ed5163f3a6cf64d789cc43734f10503db88705f/crates/network-scanner/src/lib.rs#L41-L52
- https://github.com/Devolutions/devolutions-gateway/blob/8ed5163f3a6cf64d789cc43734f10503db88705f/crates/win-api-wrappers/src/thread.rs#L33-L36

Of course, updating the code today would be massive work, so let’s just do that incrementally as we modify or add new code.

ffi/src/sspi/sec_winnt_auth_identity.rs
```diff
@@ -346,7 +349,10 @@ pub unsafe fn auth_data_to_identity_buffers_w(
 
         Ok(CredentialsBuffers::AuthIdentity(AuthIdentityBuffers {
             user,
-            domain: raw_str_into_bytes((*auth_data).domain as *const _, (*auth_data).domain_length as usize * 2),
+            domain: credentials_str_into_bytes(
+                (*auth_data).domain as *const _,
+                (*auth_data).domain_length as usize * 2,
```
CBenoit, line 354: question: do you know if domain_length is the number of code points or the number of characters of the UTF-16 string? Because some characters may be encoded on multiple code points, causing the maths to be wrong here.

CBenoit:
Thank you for improving this! I’m waiting for @pauldumais to confirm the problem is fixed on his side (discussion happening on Slack).

ffi/src/sspi/sec_handle.rs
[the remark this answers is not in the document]
TheBestTvarynka: Yes, I think we can close it

pauldumais:
> Thank you for improving this! I’m waiting for @pauldumais to confirm the problem is fixed on his side (thread on Slack).

I'm still getting the memory corruption error when I pass a blank string as username.  It seams to me that sending a blank string just like sending an invalid username like "a" should return the same failed authentication, instead of a memory exception.  I agree that null should not be supported in sspi-rs.

ffi/src/sspi/sec_handle.rs
TheBestTvarynka: done

ffi/src/sspi/sec_winnt_auth_identity.rs
TheBestTvarynka: We do not convert/reencode the string here. we only read the memory into a buffer (`Vec<u8>`). The number of bytes is `domain_length` * 2, where 2 is the size of the `unsigned short`. 
https://learn.microsoft.com/en-us/windows/win32/api/sspi/ns-sspi-sec_winnt_auth_identity_w
We use `String::from_utf16_lossy` in our codebase when we need the `String`.

So, everything should be okay here

CBenoit: I see the following at many places in the docs:

> The length, **in characters**, of the user string, not including the terminating null character.

So what you said:

> The number of **bytes** is domain_length * 2

Does not sound correct to me.

For instance, let’s take the character [𐐷](https://en.wikipedia.org/wiki/%F0%90%90%8F), which is mentioned in the [UTF-16 wikipedia page](https://en.wikipedia.org/wiki/UTF-16) as being encoded on _**2** 16-bit code points_. The string `"𐐷"` is _1_ character, but the number of bytes is 2 × 2 = _4_ which is not equal to domain_length (1 character) × 2 = 1 × 2 = 2.

Do you think you could double check that? If necessary, let’s address that in follow up PRs.

> We use `String::from_utf16_lossy` in our codebase when we need the `String`.

Note that in such cases we should probably use `String::from_utf16` instead. Otherwise, we are silently ignoring invalid code points, which should not be accepted in the first place.

CBenoit: FYI, there is a function in the crate `widestring` to handle the situation were you have a pointer to the beginning of a UTF-16 string along its length _in characters_: https://docs.rs/widestring/latest/widestring/ucstring/struct.U16CString.html#method.from_ptr

TheBestTvarynka: Hmm, that sounds reasonable. I got your idea.
Then we need to introduce a new dependency and fix the code in many places. Let's create a new issue and refactor the code in a separate PR. What do you think?

CBenoit: This sounds like the right course of action to me!

TheBestTvarynka: Let me create an issue

TheBestTvarynka: https://github.com/Devolutions/sspi-rs/issues/306

ffi/src/utils.rs
```diff
@@ -26,8 +26,26 @@ pub unsafe fn w_str_len(s: *const u16) -> usize {
...
+/// Converts raw credentials string into [Vec] of bytes.
+///
+/// Credentials are often represented as strings. For example, username, domain, password.
+/// It is OK for Windows SSPI to accept `null` or empty credential strings. The `AcquireCredentialsHandle`
+/// function will return successful status code is we pass the `null` username value. Thus, this function
+/// will return an empty [Vec] in such a case. It is done on purpose to follow the Windows SSPI behaviour.
+///
+/// # Safety
+///
+/// * `raw_buffer` must be [valid] for reads for `len` many bytes, and it must be properly aligned.
```
CBenoit, line 38: question: Is `[valid]` instead of `valid` intentional here? I don’t see any reference for that.

CBenoit:
@TheBestTvarynka Is the PR ready? I saw on Slack that the problem was fixed.

ffi/src/utils.rs
TheBestTvarynka: Nope 😅 
I ran `cargo doc` and found many other warnings. I fixed all of them (see the latest commit: https://github.com/Devolutions/sspi-rs/pull/304/commits/87fc2ab39933c63eac8868515634e24c59957ab2)

TheBestTvarynka:
> Is the PR ready?

Yes, we can merge it once CI is happy.

CBenoit (approved):
Looking good! Thank you!

PR #298, opened 2024-10-08
chore: bump Rust toolchain to 1.81.0

CBenoit:

PR #305, opened 2024-10-09
refactor: remove `lazy_static` dependency

TheBestTvarynka:
Hi,
I saw that you updated the Rust version in #298 . So, I made a small refactoring and removed the `lazy_static` dependency. We don't need it because the `LazyStatic` is stable now.


CBenoit (approved):
Good stuff, thank you!

CBenoit:
@TheBestTvarynka There is a conflict, but I approved and enabled auto-merge

PR #307, opened 2024-10-15
chore(deps): introduce workspace dependencies;

TheBestTvarynka:
Hi,
We added new dependencies to the `sspi-ffi` crate in previous PRs and I decided to finally introduce workspace dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ rsa = { version = "0.9.6", default-features = false }
 windows-sys = { version = "0.59" }
 base64 = "0.22"
 whoami = "1.5"
-tracing-subscriber = "0.3"
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
 proptest = "1.5"
 
 [features]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -29,6 +29,30 @@ exclude = [
...
+# `rt-multi-thread` feature is required for `tokio::task::block_in_place` function.
+tokio = { version = "1.40", features = ["time", "rt", "rt-multi-thread"] }
+ffi-types = { path = "crates/ffi-types" }
+winscard = { path = "crates/winscard" }
+rsa = { version = "0.9.6", default-features = false }
+windows-sys = { version = "0.59" }
+base64 = "0.22"
+whoami = "1.5"
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+proptest = "1.5"
```
CBenoit, line 54: question: Is it possible to keep the `default-features = false` and `features = […]` in the individual Cargo.toml? This is typically stuff that we want to determine at the crate level, not the workspace level. Spreading the definitions across many files is also not very good for the readability I think. I would prefer we only keep version numbers in the workspace manifest.

TheBestTvarynka: Yes, it's possible.
I didn't think about it, but now it sounds reasonable. I improved the `Cargo.toml`

CBenoit (approved):
Thank you!

PR #312, opened 2024-10-23
Patches for IronRDP server

elmarco:
Adding Negotiate support for CredSspServer, and fixing an issue with the domain name.

Note that it took me a long time to figure out an issue during NTLM NTOFWv2 check. MIC was failing because we compute the wrong hash. The reason is that we lost the domain name. FreeRDP seems to have slightly different behaviour, since it only looks up the password hash from SAM, but keeps the original user/domain fields otherwise. I refrained from updating Ntlm::custom_set_auth_identity() to only change the password fields. Thus it is up to auth_data_by_user() to be implemented to keep those fields, if that makes sense.

This PR is required for [IronRDP#558.](https://github.com/Devolutions/IronRDP/pull/558)


CBenoit (approved):
Thank you!

>  FreeRDP seems to have slightly different behaviour, since it only looks up the password hash from SAM, but keeps the original user/domain fields otherwise. I refrained from updating Ntlm::custom_set_auth_identity() to only change the password fields. Thus it is up to auth_data_by_user() to be implemented to keep those fields, if that makes sense.

I would be okay with changing more to get closer to FreeRDP too, but I think your change makes sense as far as I understand.

Do you think you could rebase on master so the duplicated commit goes away and I can rebase your two commits?

CBenoit:
@elmarco sspi v0.14.1 is published!

PR #317, opened 2024-10-28
ci: update macos-12 runners to macos-14

thenextman:
The macOS 12 runner image will be removed by December 3rd, 2024


CBenoit (approved):
Thank you!

PR #326, opened 2024-11-25
Fix domain realm mapping

### LINKED ISSUE #323: get_client_principal_realm should examine krb5.conf for domain_realm remapping
Currently `get_client_principal_realm`: https://github.com/Devolutions/sspi-rs/blob/856428be06a29371308fe14e01eac88c418cbb5c/src/kerberos/client/generators.rs#L81 generates the kerberos realm by examining username@domain and assumes domain == realm. 

krb5.conf specifies the [domain_realm] section that allows for a given domain to have a different kerberos realm eg:
`
[domain_realm]
example.com = REALM.EXAMPLE.COM
`
`get_client_principal_realm` should either examine a krb5.conf to determine the correct realm, or have some other mechanism to specify the correct realm (environment variable perhaps?)

This would solve https://github.com/gravitational/teleport/issues/49173

TheBestTvarynka:
Hi,
I've improved the domain realm mapping in this PR. Previously, we used to derive realm from the domain, but this approach sometimes results in an invalid realm.
I changed this behavior, and we are first trying to read the `krb5.conf` file `domain_realm` section. If the appropriate domain realm mapping is not found, it will continue with the default behavior which is basically `domain.to_upper_case()`.

Additionally, I wrote some tests to check it. If you have some unusual Kerberos configurations, let me know and I will add them to the tests.

closes #323

### DIFF
--- a/src/kerberos/client/generators.rs
+++ b/src/kerberos/client/generators.rs
@@ -1,3 +1,5 @@
+use std::env;
+use std::path::Path;
 use std::str::FromStr;
 
 use bitflags;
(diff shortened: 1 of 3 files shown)


CBenoit (approved):
LGTM!

PR #343, opened 2025-01-10
Add `make_signature` and `verify_signature` to `Sspi` trait.

### LINKED ISSUE #338: Supporting no-encryption NTLM message signing & GSS MIC calculation
Hello,
In an attempt to utilize the NTLM SSPI provider for signing a GSS-API negotiation message, I encountered an issue. According to the GSS-API RFC ([here](https://www.rfc-editor.org/rfc/rfc4178.html#section-5)), when the underlying scheme supports MIC calculation, it is mandatory to provide the `mechListMIC`. In my case, since NTLM was required to provide integrity, the `mechListMIC` had to be included.
I attempted to call the `Ntlm::encrypt_message` function to sign the `mechTypeList` value, as per the RFC’s requirements. However, the validation of the `mechListMIC` failed. Upon examining the implementation of the `encrypt_message` function, it became evident that it encrypts the input data before signing the digest of the data using the same RC4 state. Consequently, the RC4 state is not the initial state when signing the data. Additionally, there is an extraneous and irrelevant outcome of generating encrypted data for the `mechTypeList` data.
To address this issue, I devised a proof-of-concept (PoC) solution to sign the MIC message. I proceeded by commenting out the message encryption section from the [`encrypt_message` function](https://github.com/Devolutions/sspi-rs/blob/sspi-v0.15.0/src/ntlm/mod.rs#L395). This modification enabled the signing process to proceed successfully.
```rust
    fn encrypt_message(
        &mut self,
        _flags: EncryptionFlags,
        message: &mut [SecurityBuffer],
        sequence_number: u32,
    ) -> crate::Result<SecurityStatus> {
        if self.send_sealing_key.is_none() {
            self.complete_auth_token(&mut [])?;
        }
        println!("Starting the encryption process");
        SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Token)?; // check if exists
        let data = SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Data)?;
        let digest = compute_digest(&self.send_signing_key, sequence_number, data.data())?;

        // let encrypted_data = self.send_sealing_key.as_mut().unwrap().process(data.data());
        // if encrypted_data.len() < data.buf_len() {
        //     return Err(Error::new(ErrorKind::BufferTooSmall, "The Data buffer is too small"));
        // }
        // data.write_data(&encrypted_data)?;

        let checksum = self
            .send_sealing_key
            .as_mut()
            .unwrap()
            .process(&digest[0..SIGNATURE_CHECKSUM_SIZE]);

        let signature_buffer = SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Token)?;
        if signature_buffer.buf_len() < SIGNATURE_SIZE {
            return Err(Error::new(ErrorKind::BufferTooSmall, "The Token buffer is too small"));
        }
        let signature = compute_signature(&checksum, sequence_number);
        signature_buffer.write_data(signature.as_slice())?;

        Ok(SecurityStatus::Ok)
    }
```
Certainly, attempting to utilize the code for actual encryption of data will be futile, as that is not my current requirement. I merely need a signing mechanism as a preliminary step.
Scapy incorporates the `Gss_GstMICEx` function as part of its [NTLM module](https://github.com/secdev/scapy/blob/a8583a5132165a191d7d53ab5fdf1b6a91ea6dd9/scapy/layers/ntlm.py#L1276), which is also utilized in the encryption process (due to the shared logic).
Another noteworthy aspect is the necessity to “back up” the original send key state prior to signing the `mechTypeList` and restoring it immediately afterward (as specified in [MS-SPNG 3.3.5.1](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-spng/b87587b3-9d72-4027-8131-b76b5368115f)).
Do you have any insights regarding message signing within the NTLM SSPI (or, alternatively, in other SSPIs as well)? Specifically, would it be feasible to add such a function to the `Ntlm` implementation or, at the very least, provide a mechanism for direct access to the send key and its state, suitable for such purposes?
Thank you for your consideration.

AvivNaaman:
resolves #338

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -602,6 +602,19 @@ where
         sequence_number: u32,
     ) -> Result<SecurityStatus>;
 
+    /// TODO
+    fn make_signature(&mut self,
+        flags: SignatureFlags,
+        message: &mut [SecurityBuffer],
+        sequence_number: u32,
+    ) -> crate::Result<SecurityStatus>;
+
+    fn verify_signature(&mut self,
+        flags: SignatureFlags,
+        message: &mut [SecurityBuffer],
+        sequence_number: u32,
+    ) -> crate::Result<SignatureFlags>;
+
     /// Decrypts a message. Some packages do not encrypt and decrypt messages but rather perform and check an integrity hash.
     ///
     /// # Parameters
@@ -1063,6 +1076,13 @@ bitflags! {
     pub struct DecryptionFlags: u32 {
         const SIGN_ONLY = 0x8000_0000;
         const WRAP_NO_ENCRYPT = 0x8000_0001;
+    }
+}
+
+bitflags! {
+    /// TODO
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct SignatureFlags: u32 {
     }
 }
 
--- a/src/negotiate.rs
+++ b/src/negotiate.rs
@@ -374,6 +374,22 @@ impl Sspi for Negotiate {
             self.change_password(&mut yield_point, change_password).await
         }))
     }
+    
+    fn make_signature(&mut self,
+        flags: crate::SignatureFlags,
+        message: &mut [SecurityBuffer],
+        sequence_number: u32,
+    ) -> crate::Result<SecurityStatus> {
+        todo!()
+    }
+    
+    fn verify_signature(&mut self,
+        flags: crate::SignatureFlags,
+        message: &mut [SecurityBuffer],
+        sequence_number: u32,
+    ) -> crate::Result<crate::SignatureFlags> {
+        todo!()
+    }
 }
 
 impl SspiImpl for Negotiate {
--- a/src/ntlm/mod.rs
+++ b/src/ntlm/mod.rs
@@ -21,8 +21,8 @@ use crate::{
     ClientRequestFlags, ClientResponseFlags, ContextNames, ContextSizes, CredentialUse, DecryptionFlags,
     EncryptionFlags, Error, ErrorKind, FilledAcceptSecurityContext, FilledAcquireCredentialsHandle,
     FilledInitializeSecurityContext, InitializeSecurityContextResult, OwnedSecurityBuffer, PackageCapabilities,
-    PackageInfo, SecurityBuffer, SecurityBufferType, SecurityPackageType, SecurityStatus, ServerResponseFlags, Sspi,
-    SspiEx, SspiImpl, PACKAGE_ID_NONE,
+    PackageInfo, SecurityBuffer, SecurityBufferType, SecurityPackageType, SecurityStatus, ServerResponseFlags,
+    SignatureFlags, Sspi, SspiEx, SspiImpl, PACKAGE_ID_NONE,
 };
 
 pub const PKG_NAME: &str = "NTLM";
@@ -368,6 +368,50 @@ impl Ntlm {
             expiry: None,
         })
     }
+
+    fn compute_checksum(
+        &mut self,
+        message: &mut [SecurityBuffer],
+        sequence_number: u32,
+        digest: &[u8; 16],
+    ) -> crate::Result<SecurityStatus> {
+        let checksum = self
+            .send_sealing_key
+            .as_mut()
+            .unwrap()
+            .process(&digest[0..SIGNATURE_CHECKSUM_SIZE]);
+
+        let signature_buffer = SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Token)?;
+        if signature_buffer.buf_len() < SIGNATURE_SIZE {
+            return Err(Error::new(ErrorKind::BufferTooSmall, "The Token buffer is too small"));
+        }
+        let signature = compute_signature(&checksum, sequence_number);
+        signature_buffer.write_data(signature.as_slice())?;
+
+        Ok(SecurityStatus::Ok)
+    }
+
+    fn check_signature(
+        &mut self,
+        sequence_number: u32,
+        digest: &[u8; 16],
+        signature: &[u8],
+    ) -> crate::Result<()> {
+        let checksum = self
+            .recv_sealing_key
+            .as_mut()
+            .unwrap()
+            .process(&digest[0..SIGNATURE_CHECKSUM_SIZE]);
+        let expected_signature = compute_signature(&checksum, sequence_number);
+
+        if signature != expected_signature.as_ref() {
+            return Err(Error::new(
+                ErrorKind::MessageAltered,
+                "Signature verification failed, something nasty is going on!",
+            ));
+        }
+        Ok(())
+    }
 }
 
 impl Sspi for Ntlm {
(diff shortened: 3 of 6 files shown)


src/negotiate.rs
```diff
@@ -374,6 +374,22 @@ impl Sspi for Negotiate {
             self.change_password(&mut yield_point, change_password).await
         }))
     }
+    
+    fn make_signature(&mut self,
+        flags: crate::SignatureFlags,
+        message: &mut [SecurityBuffer],
+        sequence_number: u32,
+    ) -> crate::Result<SecurityStatus> {
+        todo!()
```
TheBestTvarynka, line 383: You can implement these methods for the negotiation module in the following way:
https://github.com/Devolutions/sspi-rs/blob/a8272f7497cff000bbed5a2a6f369b2d054adb72/src/negotiate.rs#L301-L313

src/lib.rs
```diff
@@ -602,6 +602,19 @@ where
         sequence_number: u32,
     ) -> Result<SecurityStatus>;
 
+    /// TODO
+    fn make_signature(&mut self,
+        flags: SignatureFlags,
+        message: &mut [SecurityBuffer],
+        sequence_number: u32,
+    ) -> crate::Result<SecurityStatus>;
```
TheBestTvarynka, line 610:
```suggestion
    ) -> crate::Result<()>;
```

You should not use `SecurityStatus` as a return value. The `SecurityStatus` should be used only for informing the caller about the current status of the authentication. See the following table: https://learn.microsoft.com/en-us/windows/win32/secauthn/initializesecuritycontext--kerberos#return-value.
According to the documentation (https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-makesignature), the `MakeSignature` function returns the error code (including `SEC_E_OK`). So, `Result<()>` will be enough.

src/ntlm/mod.rs
```diff
@@ -368,6 +368,50 @@ impl Ntlm {
...
+    ) -> crate::Result<SecurityStatus> {
+        let checksum = self
+            .send_sealing_key
+            .as_mut()
+            .unwrap()
+            .process(&digest[0..SIGNATURE_CHECKSUM_SIZE]);
+
+        let signature_buffer = SecurityBuffer::find_buffer_mut(message, SecurityBufferType::Token)?;
+        if signature_buffer.buf_len() < SIGNATURE_SIZE {
+            return Err(Error::new(ErrorKind::BufferTooSmall, "The Token buffer is too small"));
```
TheBestTvarynka, line 386:
```suggestion
            return Err(Error::new(ErrorKind::BufferTooSmall, "the Token buffer is too small"));
```

I know you moved this code from the `encrypt_message` function. It was written a long time ago and we are now following the new style guide (it is undocumented, but you can use this one as a reference: https://github.com/Devolutions/IronRDP/blob/master/STYLE.md#formatting-of-error-messages).

src/ntlm/mod.rs
```diff
@@ -368,6 +368,50 @@ impl Ntlm {
...
+            .recv_sealing_key
+            .as_mut()
+            .unwrap()
+            .process(&digest[0..SIGNATURE_CHECKSUM_SIZE]);
+        let expected_signature = compute_signature(&checksum, sequence_number);
+
+        if signature != expected_signature.as_ref() {
+            return Err(Error::new(
+                ErrorKind::MessageAltered,
+                "Signature verification failed, something nasty is going on!",
```
TheBestTvarynka, line 410:
```suggestion
                "signature verification failed, something nasty is going on",
```

src/lib.rs
```diff
@@ -1066,6 +1079,13 @@ bitflags! {
     }
 }
 
+bitflags! {
+    /// TODO
+    #[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+    pub struct SignatureFlags: u32 {
+    }
+}
```
TheBestTvarynka, line 1087: According to the documentation (https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-makesignature), these flags are package-specific:

> Package-specific flags that indicate the quality of protection.

So, you can use `u32` for the `pfQOP` parameter in the `make/verify_signature` methods. If you want to define some NTLM-specific flags, it's better to do it in the `ntlm` module.

TheBestTvarynka (commented):
Hi,
Thank you for your contribution! I reviewed the code and left a few comments. Overall, your code is good 👍

AvivNaaman:
@TheBestTvarynka What tests are you suggesting to add to the new functions?
Most of the current tests for encryption/decryption provide a descent validation to the signature functions (indirectly -- I suppose we'd like to add some tests for those functions specifically.
Should I copy the existing tests and modify them (to perform, for example, `make_signature` instead of `encrypt_message`, and test the output), write a whole other set of tests, or modify the existing tests to check both flows somehow?

TheBestTvarynka:
> Should I copy the existing tests and modify them (to perform, for example, make_signature instead of encrypt_message, and test the output), write a whole other set of tests, or modify the existing tests to check both flows somehow?

The signing and verification overlap with encryption and decryption a lot, so I think there is no point in testing corner cases.
I suggest you add one signing-verification test to show how it works. This will be enough.

AvivNaaman:
@TheBestTvarynka I added a test for `make_signature` validated properly by `verify_signature`, and invalid signature declined by `verify_signature`.
As far as I know, the PR is ready to merged. Let me know if any more changes are necessary to complete it.
Thanks!

TheBestTvarynka (approved):
Good code 👍 

Please, format the code (`cargo +nightly fmt --all`) and fix `clippy` warnings (`cargo clippy --all-features --all-targets`)

TheBestTvarynka:
@CBenoit the CI is happy now

CBenoit (approved):
Thank you both! Merging now :)

AvivNaaman:
Thanks for all your help @TheBestTvarynka @CBenoit!

AvivNaaman:
@CBenoit @TheBestTvarynka Regarding the changes in this pull request, do you have any expectations about releasing the next version of this crate?
I was actually about to publish the first version of my [smb-rs](https://github.com/AvivNaaman/smb-rs) crate, and cargo won't let me do it just yet, because the changes we added are "not published" to `crates.io`.
No rush or anything, because I wouldn't use it for production anyhow just yet, mostly interested.
Thanks!

CBenoit:
> Regarding the changes in this pull request, do you have any expectations about releasing the next version of this crate? I was actually about to publish the first version of my [smb-rs](https://github.com/AvivNaaman/smb-rs) crate, and cargo won't let me do it just yet, because the changes we added are "not published" to `crates.io`. No rush or anything, because I wouldn't use it for production anyhow just yet, mostly interested. Thanks!

@AvivNaaman Thank you for the reminder. I’ll try to get something out today! If not, expect it for tomorrow!

CBenoit:
Published via https://github.com/Devolutions/sspi-rs/pull/360

AvivNaaman:
@CBenoit Thanks, again!

CBenoit:
> Thanks, again!

My pleasure, thank you for contributing!

By the way, I want to add a link to your repository to our list of projects using sspi-rs: https://github.com/Devolutions/sspi-rs?tab=readme-ov-file#projects-using-sspi-rs

Is it okay for you?

PR #350, opened 2025-01-17
feat(dpapi): implement encryption and key derivation functions

TheBestTvarynka:
Hi,
I implemented encryption/decryption and key derivation functions in this PR.

The `crypto` module contains many _magic_ numbers. I took them from the Python DPAPI implementation: https://github.com/jborean93/dpapi-ng/blob/main/src/dpapi_ng/_crypto.py and https://github.com/jborean93/dpapi-ng/blob/main/src/dpapi_ng/_gkdi.py.

Docs & references:

* [[MS-GKDI]: Group Key Distribution Protocol](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-gkdi/943dd4f6-6b80-4a66-8594-80df6d2aad0a).
* [jborean93/dpapi-ng](https://github.com/jborean93/dpapi-ng/tree/main).


crates/dpapi/Cargo.toml
```diff
@@ -22,6 +22,20 @@ picky-asn1.workspace = true
 picky-asn1-der.workspace = true
 picky-asn1-x509 = { workspace = true, features = ["pkcs7"] }
 num-bigint-dig.workspace = true
+sha1.workspace = true
+sha2.workspace = true
+rand.workspace = true
+hmac.workspace = true
+
+rust-kbkdf = { version = "1.1", git = "https://gitlab.com/TheBestTvarynka/rust-kbkdf.git", branch = "fix-key-generation" }
```
TheBestTvarynka, line 30: Is it awful? Absolutely yes. 
Do we have any alternative? Not really. See this: https://github.com/RustCrypto/KDFs/issues/75#issuecomment-2444210825 

I think the best solution is to contribute to RustCrypto. Or we can fork the `rust-kbkdf` crate

CBenoit: The best approach is typically to contribute to the upstream project, because this gives back to the open source community while removing maintenance burden from our shoulders. **We definitely can’t release a crate with a git dependency.** As you said, either we contribute back, either we go the fork route with a crate we maintain and push to crates.io. I’ll link this to Marc-André, and we’ll see how we prioritize this work.

CBenoit: @TheBestTvarynka UPD: We agreed with Marc that the best course of action was to contribute KBKDF to RustCrypto.

1. Take over the [work in progress work in this PR](https://github.com/RustCrypto/KDFs/pull/87). Collaborate with the author and RustCrypto maintainers to implement the missing pieces so that PR can be merged. Check with them if you can open a new PR from your fork, or work from theirs.
2. Let RustCrypto publish the crate and manage it.
3. Remove the git dependency from dpdpi in favor of the proper crates.io dependency.
4. Integrate and publish our dpdpi crate.

Your priority should be to move this forward so we’re not stuck with a git dependency when it’s time to integrate dpdpi into sspi.
Feel free to rotate back on the dpdpi work when you wait for their feedback.
RustCrypto maintainers are pretty responsive, so I believe things should move forward quickly 🙂

TheBestTvarynka: Great!

CBenoit: And, it’s fine if we merge this PR, even with the git dependency, because dpdpi is not depended on by our other crates that we are publishing to crates.io yet.

CBenoit (approved):
LGTM!

PR #354, opened 2025-01-20
fix: store session key when using Ntlm as a server

### LINKED ISSUE #349: `struct Ntlm` field `session_key` never set to non-`None` value
`struct Ntlm` has a `session_key` field which is `Option<[u8; SESSION_KEY_SIZE]>`:

https://github.com/Devolutions/sspi-rs/blob/a8272f7497cff000bbed5a2a6f369b2d054adb72/src/ntlm/mod.rs#L85-L94

In all cases this field seems to be initialised with the value `None`. However, I don't see anywhere in the code where this is set to a non-`None` value. As such the getter for the `session_key` field always returns `None`:

https://github.com/Devolutions/sspi-rs/blob/a8272f7497cff000bbed5a2a6f369b2d054adb72/src/ntlm/mod.rs#L214-L216

It looks like the place where this should be set is:

https://github.com/Devolutions/sspi-rs/blob/a8272f7497cff000bbed5a2a6f369b2d054adb72/src/ntlm/messages/server/complete_authenticate.rs#L36-L58

E.g. a `context.session_key = Some(session_key);` line before, say, line 55 `context.state = NtlmState::Final;`

malcolmstill:
Fixes #349 by storing session key when `fn complete_authenticate` is called

### DIFF
--- a/src/ntlm/messages/server/complete_authenticate.rs
+++ b/src/ntlm/messages/server/complete_authenticate.rs
@@ -52,6 +52,7 @@ pub fn complete_authenticate(context: &mut Ntlm) -> crate::Result<SecurityStatus
         session_key.as_ref(),
     )?;
 
+    context.session_key = Some(session_key);
     context.state = NtlmState::Final;
 
     Ok(SecurityStatus::Ok)


TheBestTvarynka (approved):
LGTM
Thank you 🙏

Issue #120, opened 2023-04-24
Support SECBUFFER_READONLY_WITH_CHECKSUM in EncryptMessage and DecryptMessage


jborean93:
In order to support RPC message encryption and signing, the `EncryptMessage` and `DecryptMessage` needs to support security buffers marked as `SECBUFFER_READONLY_WITH_CHECKSUM`. This flag indicates the bytes in that buffer are not part of the encrypted payload but are included in the header signature. For example here is what the SSPI calls look like by the RPC client there:

```
EncryptMessage(Context: 0x1E752D17F28, Qop: 0, Message: 0xC8EED8D878, SeqNo: 2)
        EncryptMessage Message(Version: 0, Buffers: 5)
                [0] Type: SECBUFFER_DATA | SECBUFFER_READONLY_WITH_CHECKSUM (1), Length: 24, Data: 05000003100000004C014C0008000000D800000001000000
                [1] Type: SECBUFFER_DATA (1), Length: 224, Data: 6C000000000000006C000000000000000100048054000000600000000000000014000000020040000200000000002400030000000105000000000005150000001D9377F744357ACC8CD37BA9000200000000140002000000010100000000000100000000010100000000000512000000010100000000000512000000000000000000020000000000A84FC6BA90E87C91109083E7B0F859966901000010000000020000008AE3137102F4367102402800605978B94F52DF118B6D83DCDED720850100000033057171BABE37498319B5DBEF9CCC36010000000000000000000000
                [2] Type: SECBUFFER_DATA | SECBUFFER_READONLY_WITH_CHECKSUM (1), Length: 8, Data: 0906080000000000
                [3] Type: SECBUFFER_TOKEN (2), Length: 76, Data: 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
                [4] Type: SECBUFFER_PKG_PARAMS | SECBUFFER_READONLY (3), Length: 12, Data: 020000000000000000EBBFE8
EncryptMessage -> Res: 0x00000000
        EncryptMessage Message(Version: 0, Buffers: 5)
                [0] Type: SECBUFFER_DATA | SECBUFFER_READONLY_WITH_CHECKSUM (1), Length: 24, Data: 05000003100000004C014C0008000000D800000001000000
[...]

PR #357, opened 2025-02-04
feat(sspi): add  `SECBUFFER_READONLY_WITH_CHECKSUM` flag support

TheBestTvarynka:
Hi,
I added `SECBUFFER_READONLY` and `SECBUFFER_READONLY_WITH_CHECKSUM` flags support. It is related to https://github.com/Devolutions/sspi-rs/issues/120

This PR affects many files because I changed the security buffer structure to add those flags.

Docs & reference:

* [`SecBuffer` structure (sspi.h)](https://learn.microsoft.com/en-us/windows/win32/api/sspi/ns-sspi-secbuffer).


Cargo.toml
```diff
@@ -37,11 +37,11 @@ sha1 = { version = "0.10", default-features = false }
...
-picky = "7.0.0-rc.11"
-picky-asn1 = "0.10"
-picky-asn1-der = "0.5"
-picky-asn1-x509 = "0.14"
-picky-krb = "0.9"
+picky = { path = "../picky-rs/picky" }
+picky-asn1 = { path = "../picky-rs/picky-asn1" }
+picky-asn1-der = { path = "../picky-rs/picky-asn1-der" }
+picky-asn1-x509 = { path = "../picky-rs/picky-asn1-x509" }
+picky-krb = { path = "../picky-rs/picky-krb" }
```
TheBestTvarynka, line 44: The PR depends on changes in picky-rs: https://github.com/Devolutions/picky-rs/pull/342

CBenoit: https://github.com/Devolutions/picky-rs/pull/343

src/security_buffer.rs
```diff
@@ -26,82 +19,224 @@ pub enum SecurityBuffer<'data> {
     Empty,
 }
 
+/// A special security buffer type is used for the data decryption. Basically, it's almost the same
+/// as `OwnedSecurityBuffer` but for decryption.
```
CBenoit, line 23: thought: What about `SecurityBuffer` for the owned version, and `SecurityBufferRef` for the borrowing version? It’s a common naming I’m seeing more and more. (Follow up in a separate PR.)

src/security_buffer.rs
```diff
@@ -1,20 +1,13 @@
...
+/// A security buffer type with a mutable reference to the buffer data.
 ///
-/// So, the already defined `OwnedSecurityBuffer` is not suitable for decryption because it uses [Vec] inside.
-/// We use reference in the [SecurityBuffer] structure to avoid data cloning as much as possible.
-/// Decryption/encryption input buffers can be very large. Even up to 32 KiB if we are using this crate as a TSSSP(CREDSSP)
-/// security package.
+/// Basically, it is a security buffer but without buffer flags.
 #[non_exhaustive]
-pub enum SecurityBuffer<'data> {
+pub enum SecurityBufferType<'data> {
```
CBenoit, line 10: naming: What do you think of `UntaggedSecurityBuffer`? Does it make sense in this context?

src/security_buffer.rs
```diff
@@ -110,29 +245,64 @@ impl<'data> SecurityBuffer<'data> {
...
+        buffers.iter().find(|b| b.buffer_type() == buffer_type).ok_or_else(|| {
+            Error::new(
+                ErrorKind::InvalidToken,
+                format!("no buffer was provided with type {:?}", buffer_type),
+            )
+        })
+    }
+
+    /// Returns the vector of immutable references to the [SecurityBuffer] with specified buffer type.
+    pub fn buffers_with_type<'a>(
```
CBenoit, line 259: naming: `of` instead of `with` is clearer. Same suggestion for the other filtered iterators.
```suggestion
    pub fn buffers_of_type<'a>(
```

CBenoit (commented):
Nice work!

src/security_buffer.rs
TheBestTvarynka: Good idea! Your naming suggestions are good as always. Somewhat I didn't think about the `-Ref` postfix.
Done

src/security_buffer.rs
TheBestTvarynka: Actually, it is tagged because each enum variant represents a different buffer type. It lacks only buffer flags, so I renamed it to `UnflaggedSecurityBuffer`. What do you think?

src/security_buffer.rs
TheBestTvarynka: done

Cargo.toml
TheBestTvarynka: done

CBenoit (approved):
LGTM!

PR #368, opened 2025-02-18
feat(dpapi): implement RPC client and authentication

TheBestTvarynka:
Hi,
In this PR, I implemented RPC client and RPC authentication and added some comments to the code to make it easier to understand.

Docs & references:

* [The Kerberos Version 5 Generic Security Service Application Program Interface (GSS-API)](https://www.rfc-editor.org/rfc/rfc4121).
* [SecBuffer structure (sspi.h)](https://learn.microsoft.com/en-us/windows/win32/api/sspi/ns-sspi-secbuffer).
* [[MS-RPCE]: Remote Procedure Call Protocol Extensions](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-rpce/290c38b1-92fe-4229-91e6-4fc376610c15).


crates/dpapi/src/lib.rs
```diff
@@ -9,4 +12,4 @@ pub mod rpc;
 pub(crate) mod sid;
 pub(crate) mod str;
 
-pub use error::*;
+pub use error::{DpapiResult, Error};
```
CBenoit, line 15: note: Naming is a bit inconsistent.

`DpapiResult` / `Error` instead of 1) `DpapiResult` / `DpapiError` or 2) `Result` / `Error`.

Both 1) and 2) are equally good, but I strongly suggest to prefix with the crate name (without importing) for 2): `dpapi::Result` / `dpapi::Error`

crates/dpapi/src/rpc/auth.rs
```diff
@@ -0,0 +1,234 @@
...
+#[derive(Error, Debug)]
+pub enum AuthError {
+    #[error("{0} security provider is not supported")]
+    SecurityProviderNotSupported(&'static str),
+
+    #[error("SSPI authorization error: {0}")]
+    Sspi(#[from] sspi::Error),
+
+    #[error(transparent)]
+    IntConversion(#[from] std::num::TryFromIntError),
```
CBenoit, line 20: suggestion: Note that at this level I think it’s more relevant to give a high-level meaning to the failed conversion.

Example: `BadSecurityTrailer`, without the from-conversion.

This `IntConversion` is convenient and straightforward, but does not tell much of the story.

crates/dpapi/src/rpc/verification.rs
```diff
@@ -0,0 +1,259 @@
...
+    pub flags: CommandFlags,
+}
+
+impl CommandBitmask {
+    fn from_flags_and_value(flags: CommandFlags, value: &[u8]) -> DpapiResult<Self> {
+        if value.len() != 4 {
+            Err(CommandError::InvalidCommandBitmaskValueLength(value.len()))?;
+        }
+
+        let bits: [u8; 4] = value.try_into().expect("length should be checked");
```
CBenoit, line 125: nitpick: A clearer wording is:
```suggestion
        let bits: [u8; 4] = value.try_into().expect("length is checked above");
```

CBenoit (approved):
LGTM! Feel free to merge yourself.

crates/dpapi/src/lib.rs
TheBestTvarynka: I prefer the second option :)
done

crates/dpapi/src/rpc/auth.rs
TheBestTvarynka: done

crates/dpapi/src/rpc/verification.rs
TheBestTvarynka: done

PR #371, opened 2025-02-19
feat(dpapi): replace `rust-kbkdf` with `kbkdf` crate

TheBestTvarynka:
Hi,
I replaced the `rust-kbkdf` with the `kbkdf` crate in this PR.


crates/dpapi/Cargo.toml
```diff
@@ -29,7 +29,12 @@ rand.workspace = true
 hmac.workspace = true
 tracing = { workspace = true, default-features = true }
 
-rust-kbkdf = { version = "1.1", git = "https://gitlab.com/TheBestTvarynka/rust-kbkdf.git", branch = "fix-key-generation" }
+kbkdf = "0.0.1"
+sha1-pre = { version = "0.11.0-pre.2", package = "sha1" }
+sha2-pre = { version = "0.11.0-pre.2", package = "sha2" }
+hmac-pre = { version = "0.13.0-pre.4", package = "hmac" }
+digest-pre = { version = "0.11.0-pre.9", package = "digest", default-features = false }
```
TheBestTvarynka, line 36: The `kbkdf` crate depends on the latest versions of the `sha*` and `digest` crates. I tried older ones, but I got only compile errors about unsatisfied trait bounds.
I cannot update workspace dependencies to the latest version. I tried it too and got a lot of compilation errors in `sspi` and `picky-*` crates (I'm not willing to fix them now. let's do it in another PR)

crates/dpapi/src/crypto/mod.rs
```diff
@@ -174,35 +179,38 @@ pub fn content_encrypt(
...
+        ($prf_ty:ty, { $($l_value:expr => $l_ty:ty,)* }) => {
+            $(
+                if $l_value == length {
+                    return Ok(Counter::<$prf_ty, $l_ty>::default()
+                        .derive(Params::builder(secret).with_label(label).with_context(context).use_counter(true).use_l(true).build())?
+                        .to_vec());
+                }
+            )*
+        };
     }
```
TheBestTvarynka, line 196: Key length is a type parameter in the `kbkdf` implementation, but our key length is a runtime value. so, we need to move from runtime value to compile time.
The only way to solve this is to check a variable value in runtime and specify the needed type parameter. I wrote this macro to simplify this process (see usage below)

crates/dpapi/src/crypto/mod.rs
```diff
@@ -174,35 +179,38 @@ pub fn content_encrypt(
...
+        HashAlg::Sha256 => {
+            derive_key!(Hmac<sha2_pre::Sha256>, { 32 => Kout32, 64 => Kout64, });
+        }
+        HashAlg::Sha384 => {
+            derive_key!(Hmac<sha2_pre::Sha384>, { 32 => Kout32, 64 => Kout64, });
+        }
+        HashAlg::Sha512 => {
+            derive_key!(Hmac<sha2_pre::Sha512>, { 32 => Kout32, 64 => Kout64, });
+        }
     }
```
TheBestTvarynka, line 211: I can generate this `match` in the macro too, but that way macro will become too complex and unreadable (example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3d37bce0287bb2a063d51cf24c48c2ae)

crates/dpapi/Cargo.toml
CBenoit: Oh okay… The next step would be to wait for these dependencies to go out of the prerelease phase so we can publish `kbkdf`. It’s fine for now, at least the migration work is done 👍

crates/dpapi/src/crypto/mod.rs
CBenoit: It’s fine to keep the match spelled out like this 👍 
It’s not like we are going to write it elsewhere.

CBenoit (approved):
LGTM! Good job getting rid of the legacy KBKDF backend so far 👏

PR #378, opened 2025-02-24
feat(dpapi-cli): implement simple DPAPI cli client;

TheBestTvarynka:
Hi,
I implemented a simple DPAPI client for secret encryption and decryption.

This app will be used for blob decryption/encryption and DPAPI implementation testing (by QAs).

### DIFF
--- a/crates/dpapi-cli-client/Cargo.toml
+++ b/crates/dpapi-cli-client/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "dpapi-cli-client"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+clap = { version = "4.5", features = ["derive"] }
+dpapi = { path = "../dpapi" }
--- a/crates/dpapi-cli-client/Cargo.toml
+++ b/crates/dpapi-cli-client/Cargo.toml
@@ -4,6 +4,6 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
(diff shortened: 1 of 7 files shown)


crates/dpapi-cli-client/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "dpapi-cli-client"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+clap = { version = "4.5", features = ["derive"] }
```
CBenoit, line 7: suggestion: Since this is just a small testing tool, could you consider using xflags instead of clap?

Clap really improved on the compilation time and the binary size, but it’s still much much bigger because of the scope size is very different.

https://lib.rs/crates/xflags + https://lib.rs/crates/xflags-macros
image image
(no syn)

https://lib.rs/crates/clap
image
(uses syn)

Typically, a small utility like xflags is more than enough. Feel free to use the candidate version, it doesn’t matter.

CBenoit (approved):
LGTM! Just one suggestion to avoid bringing many more dependencies. (That being said, I do agree that the overall impact may not be so big, we already depend on a little bit too many dependencies I believe.)

crates/dpapi-cli-client/Cargo.toml
TheBestTvarynka: done

CBenoit: Thank you!

PR #380, opened 2025-02-25
feat(ffi): implement DPAPI FFI functions

TheBestTvarynka:
Hi,
I implemented DPAPI FFI functions in this PR. I put the code under the optional `dpapi` feature (enabled by default). Additionally, I wrote a simple test that checks FFI implementation (it's better to run it with Miri).

### DIFF
--- a/ffi/src/dpapi/api.rs
+++ b/ffi/src/dpapi/api.rs
@@ -5,14 +5,15 @@ mod inner {
 
 #[cfg(any(test, miri))]
 mod inner {
-    /// We have FFI wrappers for DPAPI functions from the [dpapi] crate and we want to test them.
-    /// The DPAPI implementation is complex and makes calls to the RPC and KDC servers.
-    /// Implementing a mock for KDF and RPC servers is too hard and unreasonable. So, we wrote a simple
-    /// high-level mock of [n_crypt_protect_secret] and [n_crypt_unprotect_secret] functions.
-    ///
-    /// **Note**: The goal is to test FFI functions, not the DPAPI implementation correctness.
-    /// The FFI tests should not care about returned data correctness but rather check
-    /// for memory corruptions and memory leaks.
+    //! We have FFI wrappers for DPAPI functions from the [dpapi] crate and we want to test them.
+    //! The DPAPI implementation is complex and makes calls to the RPC and KDC servers.
+    //! Implementing a mock for KDF and RPC servers is too hard and unreasonable. So, we wrote a simple
+    //! high-level mock of [n_crypt_protect_secret] and [n_crypt_unprotect_secret] functions.
+    //!
+    //! **Note**: The goal is to test FFI functions, not the DPAPI implementation correctness.
+    //! The FFI tests should not care about returned data correctness but rather check
+    //! for memory corruptions and memory leaks.
+
     use sspi::Secret;
     use uuid::Uuid;
 
--- a/ffi/src/dpapi/mod.rs
+++ b/ffi/src/dpapi/mod.rs
@@ -270,7 +270,7 @@ pub unsafe extern "system" fn DpapiFree(buf: LpCByte) -> u32 {
     catch_panic! {
         check_null!(buf);
 
-        // SAFETY: The user should uphold that the passed pointer is a memory allocated by a out DPAPI functions.
+        // SAFETY: The user should uphold that the passed pointer is a memory allocated by an out DPAPI functions.
         unsafe {
             libc::free(buf as _);
         }
--- a/ffi/src/lib.rs
+++ b/ffi/src/lib.rs
@@ -7,7 +7,7 @@ extern crate tracing;
 
 #[cfg(feature = "dpapi")]
 #[deny(unsafe_op_in_unsafe_fn)]
-#[deny(clippy::undocumented_unsafe_blocks)]
+#[warn(clippy::undocumented_unsafe_blocks)]
 pub mod dpapi;
 pub mod logging;
 pub mod sspi;


ffi/src/dpapi/api.rs
```diff
@@ -0,0 +1,42 @@
...
+#[cfg(any(test, miri))]
+mod inner {
+    /// We have FFI wrappers for DPAPI functions from the [dpapi] crate and we want to test them.
+    /// The DPAPI implementation is complex and makes calls to the RPC and KDC servers.
+    /// Implementing a mock for KDF and RPC servers is too hard and unreasonable. So, we wrote a simple
+    /// high-level mock of [n_crypt_protect_secret] and [n_crypt_unprotect_secret] functions.
+    ///
+    /// **Note**: The goal is to test FFI functions, not the DPAPI implementation correctness.
+    /// The FFI tests should not care about returned data correctness but rather check
+    /// for memory corruptions and memory leaks.
```
CBenoit, line 15: issue: You probably want the docstring to apply to the item above (the module)
```suggestion
    //! We have FFI wrappers for DPAPI functions from the [dpapi] crate and we want to test them.
    //! The DPAPI implementation is complex and makes calls to the RPC and KDC servers.
    //! Implementing a mock for KDF and RPC servers is too hard and unreasonable. So, we wrote a simple
    //! high-level mock of [n_crypt_protect_secret] and [n_crypt_unprotect_secret] functions.
    //!
    //! **Note**: The goal is to test FFI functions, not the DPAPI implementation correctness.
    //! The FFI tests should not care about returned data correctness but rather check
    //! for memory corruptions and memory leaks.

```

ffi/src/lib.rs
```diff
@@ -5,6 +5,10 @@
 #[macro_use]
 extern crate tracing;
 
+#[cfg(feature = "dpapi")]
+#[deny(unsafe_op_in_unsafe_fn)]
+#[deny(clippy::undocumented_unsafe_blocks)]
```
CBenoit, line 10: praise: Excellent initiative!

note: I think it’s fine to use `#[warn]` instead of `#[deny]` because we deny warnings in the CI anyway.

thought: I was going to suggest that we [put these in the Cargo.toml](https://github.com/Devolutions/IronRDP/blob/2a98f8be3e92e3e3e7e91eb2b3149d8399153fef/Cargo.toml#L78-L160), but at this point I guess we would have too much to fix at once in the whole ffi crate, right? 😢 

suggestion: Other lints I suggest when dealing with unsafe code:

- invalid_reference_casting
- clippy::multiple_unsafe_ops_per_block
- clippy::transmute_ptr_to_ptr
- clippy::as_ptr_cast_mut
- clippy::cast_ptr_alignment
- clippy::fn_to_numeric_cast_any
- clippy::ptr_cast_constness

ffi/src/dpapi/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+    use std::ptr::{null, null_mut};
+
+    use super::*;
+
+    /// This test simulates `DpapiProtectSecret`, `DpapiUnprotectSecret`, and `DpapiFree` function calls.
+    /// It's better to run it using Miri: https://github.com/rust-lang/miri.
+    /// cargo +nightly miri test
+    ///
+    /// Note: this test aims to check only the implementation of FFI functions.
+    /// Checking the correctness of DPAPI functions is not a goal of this test.
```
CBenoit, line 293: praise: Miri tests, very appreciated!

ffi/src/dpapi/mod.rs
```diff
@@ -0,0 +1,349 @@
...
+///
+/// The `data` parameter must be a valid pointer to the memory allocated by the [DpapiProtectSecret] or
+/// [DpapiUnprotectSecret] functions and **cannot be NULL**.
+#[instrument(skip_all)]
+#[no_mangle]
+pub unsafe extern "system" fn DpapiFree(buf: LpCByte) -> u32 {
+    catch_panic! {
+        check_null!(buf);
+
+        // SAFETY: The user should uphold that the passed pointer is a memory allocated by a out DPAPI functions.
```
CBenoit, line 273: typo:
```suggestion
        // SAFETY: The user should uphold that the passed pointer is a memory allocated by an out DPAPI functions.
```

CBenoit (commented):
Excellent work!

> Additionally, I wrote a simple test that checks FFI implementation (it's better to run it with Miri).

That’s very good. And our CI is already running tests from the ffi crate using miri 😄

ffi/src/lib.rs
TheBestTvarynka: > note: I think it’s fine to use #[warn] instead of #[deny] because we deny warnings in the CI anyway.

done

> but at this point I guess we would have too much to fix at once in the whole ffi crate, right? 😢

right, but it's work that needs to be done. We must refactor our FFI if we want to use the 2024 edition.

> The [unsafe_op_in_unsafe_fn](https://doc.rust-lang.org/nightly/rustc/lints/listing/allowed-by-default.html#unsafe-op-in-unsafe-fn) lint now warns by default. ([src](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/unsafe-op-in-unsafe-fn.html)).

There is not much work in the `winscard` module (because some lints were enabled from the start), but a lot of refactoring in the `sspi` module

ffi/src/dpapi/mod.rs
TheBestTvarynka: done

ffi/src/dpapi/api.rs
TheBestTvarynka: Oh, yes. thanks!
done

ffi/src/lib.rs
TheBestTvarynka: > Do you prefer adding the lints I mentioned now in a separate PR?

yes. I don't want to mix changes 😅. I'd like to do it after the `dev/dpapi` branch merge

PR #395, opened 2025-03-12
fix(dpapi): NTLM RPC auth

TheBestTvarynka:
Hi,
I fixed the NTLM RPC auth in this PR. Explanation:

Our Kerberos implementation has parts of SPNEGO and GSS API tokens inside. The first phase of the Kerberos implementation is called `Negotiate`: https://github.com/Devolutions/sspi-rs/blob/bda97ded7807e15f92ad796f881f85ef27839d96/src/kerberos/mod.rs#L109 + https://github.com/Devolutions/sspi-rs/blob/bda97ded7807e15f92ad796f881f85ef27839d96/src/kerberos/mod.rs#L800-L832. And we don't need it during the RPC auth. So, we just decided to call the `initialize_security_context` method twice and just ignore the first call result: https://github.com/Devolutions/sspi-rs/blob/bda97ded7807e15f92ad796f881f85ef27839d96/crates/dpapi/src/rpc/client.rs#L350-L354.
But it works only when `Kerberos` or `Negotiate(Kerberos)` security package is used. When we tried to authorize using NTLM we got an error. Because we don't need to call the `initialize_security_context` method twice. My fix is pretty straightforward: I check for the security type in use and call the `initialize_security_context` method second time only when needed.

The second problem was in the last returned token from the `initialize_security_context` method. We don't need to return _the final neg token_ in the case of RPC Kerberos auth: https://github.com/Devolutions/sspi-rs/blob/bda97ded7807e15f92ad796f881f85ef27839d96/src/kerberos/mod.rs#L249-L267. So, I just removed it.

I retested the DPAPI implementation with NTLM and Kerberos auth. Everything works great.

### DIFF
--- a/crates/dpapi/src/rpc/client.rs
+++ b/crates/dpapi/src/rpc/client.rs
@@ -347,11 +347,18 @@ impl RpcClient {
     /// The bind/bind_ack exchange continues until authentication is finished.
     #[instrument(level = "trace", ret, skip(self))]
     pub fn bind_authenticate(&mut self, contexts: &[ContextElement]) -> Result<BindAck> {
-        // The first `initialize_security_context` call is Negotiation in our Kerberos implementation.
-        // We don't need its result in RPC authentication.
-        let _security_trailer = self.auth.initialize_security_context(Vec::new())?;
-
         let security_trailer = self.auth.initialize_security_context(Vec::new())?;
+
+        // This is a small dirty trick. We do not have separate abstractions for SPNEGO and GSS API.
+        // Thus, our Kerberos implementation contains parts of SPNEGO and GSS API. But we do not need
+        // them during the RPC auth. We skip the unneeded output by calling the `initialize_security_context`
+        // method twice.
+        let security_trailer = if self.auth.needs_negotication() {
+            self.auth.initialize_security_context(Vec::new())?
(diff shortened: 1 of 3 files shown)


crates/dpapi/src/rpc/client.rs
```diff
@@ -347,11 +347,18 @@ impl RpcClient {
...
+
+        // This is a small dirty trick. We do not have separate abstractions for SPNEGO and GSS API.
+        // Thus, our Kerberos implementation contains parts of SPNEGO and GSS API. But we do not need
+        // them during the RPC auth. We skip the unneeded output by calling the `initialize_security_context`
+        // method twice.
+        let security_trailer = if self.auth.needs_negotication() {
+            self.auth.initialize_security_context(Vec::new())?
+        } else {
+            security_trailer
+        };
```
TheBestTvarynka, line 360: We may want to refactor it in the future, but for now it's too much work.

CBenoit (approved):
LGTM! Should I cut a new release of Devolutions.Sspi?

TheBestTvarynka:
> May I cut a new release of Devolutions.Sspi now, or do you still have things to push?

Feel free to publish a new release

PR #402, opened 2025-03-18
refactor(dapapi): encoding/decoding traits

TheBestTvarynka:
Hi,
I rework the encoding/decoding traits. Our goal is to support WASM. I took your past comment (https://github.com/Devolutions/sspi-rs/pull/342#discussion_r1911565582) into account and tried to make it similar to `IronRDP` and other projects.

I didn't change any DPAPI-related logic. This PR contains only decoding/encoding refactoring.


crates/dpapi-core/Cargo.toml
CBenoit, line 1: It seems you copied a lot of code from `ironrdp-core`. Do you think it would make sense to depend on `ironrdp-core` (the crate is available on crates.io), and re-export some API using `pub use` instead? To avoid maintaining the same code at two different places.

crates/dpapi-core/src/encode.rs
```diff
@@ -0,0 +1,106 @@
...
+        let frame_length = self.frame_length();
+        let mut buf = ::alloc::vec![0; frame_length];
+        let written = self.encode(&mut buf)?;
+        debug_assert_eq!(written, frame_length);
+
+        Ok(buf)
+    }
+
+    /// Computes size in bytes required for encoding of this PDU.
+    fn frame_length(&self) -> usize;
```
CBenoit, line 44: I would reiterate this comment: https://github.com/Devolutions/sspi-rs/pull/342#discussion_r1911563909

crates/dpapi-core/Cargo.toml
TheBestTvarynka: Hm, I think we can. Let me try

crates/dpapi-core/src/encode.rs
TheBestTvarynka: Oh, right. I forgot about it. thank you!

crates/dpapi-core/src/encode.rs
```diff
@@ -0,0 +1,84 @@
...
+
+        dst.write_slice(&self.to_bytes_le());
+
+        Ok(())
+    }
+
+    fn size_ext(&self) -> usize {
+        Uuid::FIXED_PART_SIZE
+    }
+}
```
TheBestTvarynka, line 26: In general, we have two options how to implement encoding and decoding for foreign types.1

1. Introduce an extension trait (how I did)
2. Or write helper functions like `fn encode_uuid`, `fn encode_vec<T: Encode>`, etc.

These two options are equal to me. Do you have any thoughts about it?

TheBestTvarynka:
Do not merge it yet :) currently, I'm testing it on my environment after the refactoring

TheBestTvarynka:
> Do not merge it yet :) currently, I'm testing it on my environment after the refactoring

Everything works well 😊

crates/dpapi-core/src/encode.rs
CBenoit: I tend to prefer 2. unless I really want the operation as a postfix (`value.operation(…)`). I’m fine with both, so feel free to take the approach you prefer!

CBenoit: Oh, sorry I didn’t realize you were asking for "foreign types". You can’t implement a foreign trait on a foreign type because of the orphan rule. In those case, I think the typical approach is to use the new type pattern.

I believe the current approach is a bit strange. This is not an "extension trait".

An [extension trait](https://rust-lang.github.io/rfcs//0445-extension-trait-conventions.html#what-is-an-extension-trait) either adds new methods to a foreign type:

```rust
use foo::Foo;

trait FooExt {
    fn bar(&self);
}

impl FooExt for Foo {
    fn bar(&self) { .. }
}
```

Or add extra methods for all types implementing another core trait:
```rust
// The core, object-safe trait
trait Iterator<A> {
    fn next(&mut self) -> Option<A>;
}

// The extension trait offering object-unsafe methods
trait IteratorExt<A>: Iterator<A> {
    fn chain<U: Iterator<A>>(self, other: U) -> Chain<Self, U> { ... }
    fn zip<B, U: Iterator<B>>(self, other: U) -> Zip<Self, U> { ... }
    fn map<B>(self, f: |A| -> B) -> Map<'r, A, B, Self> { ... }
    ...
}

// A blanket impl
impl<A, I> IteratorExt<A> for I where I: Iterator<A> {
    ...
}
```

`EncodeExt` is none of these, and so I believe this code can be considered confusing, and goes against a few good practices on generic code.

CBenoit: If you don’t need to use the foreign types in generic functions, my suggestion would be to define plain old functions instead. What do you think?

CBenoit: If it’s really necessary, we can consider adding feature flags to ironrdp-core in order to add implementation for some types such as UUID, but I feel it’s better to use a free standing function. Especially, it’s not obvious _how_ a UUID should be encoded for various protocols, and providing a one-fits-all implementation may not be appropriate.

TheBestTvarynka: > If you don’t need to use the foreign types in generic functions, my suggestion would be to define plain old functions instead. What do you think?

yes, plain functions would be better. I just pushed a new commit. Can you take another look?

crates/dpapi-core/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+pub mod gkdi;
+mod marker;
+mod padding;
+pub mod rpc;
+pub mod str;
+
+pub use decode::{DecodeWithContextOwned, FindLength, FixedPartSize, decode_uuid, read_c_str_utf16_le};
+pub use encode::{EncodeVec, encode_seq, encode_uuid, size_seq};
+pub use error::{Error, Result};
+pub use ironrdp_core as core;
```
CBenoit, line 20: suggestion: Instead of re-exporting ironrdp_core entirely, re-export the traits and other types you actually need only. E.g.: `Encode`/`Decode` traits.

crates/dpapi-core/src/padding.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub fn read(len: usize, src: &mut ReadCursor<'_>) -> DecodeResult<()> {
+        let padding_len = Self::padding(len);
+        ensure_size!(ctx: "Padding", in: src, size: padding_len);
+
+        src.advance(padding_len);
+
+        Ok(())
+    }
+
+    pub fn padding(len: usize) -> usize {
```
CBenoit, line 25: Maybe it can be constified?
```suggestion
    pub const fn padding(len: usize) -> usize {
```

crates/dpapi-core/src/lib.rs
TheBestTvarynka: done

crates/dpapi-core/src/padding.rs
TheBestTvarynka: done

crates/dpapi-core/src/padding.rs
CBenoit, line 1: Small word of caution about the `write` function: it’s typically better to ensure the buffer contains zeroes by actually writing zeroes. Especially important when reusing the write buffer, to ensure we are not sending leftover bytes previously written.

Also, I don’t see any real benefit to using a struct with a const generic parameter for the alignment here.
`ALIGNMENT` is used like any regular parameter, but with an awkward syntax (turbofish).

The methods should probably be converted into a free standing functions such as:
```rust
pub const fn compute_padding(alignment: usize, len: usize) -> usize {
    (alignment - (len % alignment)) % alignment
}
```

Note that when alignment is known at compile-time, the compiler will generate equally performant code (as an optimization). This function is also usable in any const context.

As an aside, IronRDP also comes with some helpers for the padding that you may use as building blocks.
For some reason the padding helpers of IronRDP were not in the right crate.
I opened a PR to move them into ironrdp-core: https://github.com/Devolutions/IronRDP/pull/716
The purpose is slightly different: there is no alignment parameter. But you may reuse the write_padding function, once you computed the required padding using the alignment and the total length.

(PS: I do like the turbofish operator, but I would rather not write it too much :joy:)

crates/dpapi-core/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+version = "0.1.0"
+edition = "2024"
+
+[features]
+std = []
+
+[dependencies]
+bitflags.workspace = true
+num-derive.workspace = true
+num-traits = { workspace = true, default-features = true }
```
CBenoit, line 12: note: I typically recommend against using `num-derive` and `num-traits` crates.

Two main reasons:

- [For resilient parsing, we should typically avoid enums](https://github.com/Devolutions/IronRDP/tree/b72e0857bfe2bcf502db62e08da605412d9ba80c/crates/ironrdp-pdu#enumeration-like-types-should-allow-resilient-parsing), and
- Otherwise, [it’s typically not worth adding a proc-macro in a core crate](https://github.com/Devolutions/IronRDP/blob/b72e0857bfe2bcf502db62e08da605412d9ba80c/ARCHITECTURE.md#core-tier). (This also applies to `thiserror`.)

crates/dpapi-core/src/decode.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/// Finds the precise byte count required to decode the frame from a possibly partial input.
+/// Incorrectly reading too few or too many bytes will lead to program malfunctions.
+pub trait FindLength: FixedPartSize {
+    /// Try to find the length of this frame given the first bytes.
+    fn find_frame_length(bytes: &[u8]) -> DecodeResult<Option<usize>>;
+}
+
+pub fn read_c_str_utf16_le(len: usize, src: &mut ReadCursor<'_>) -> DecodeResult<String> {
```
CBenoit, line 38: question: Should this be moved to the `str` module?

crates/dpapi-core/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+
+extern crate alloc;
+
+mod decode;
+mod encode;
+mod error;
+pub mod gkdi;
+mod marker;
+mod padding;
+pub mod rpc;
```
CBenoit, line 14: note: It seems `gkdi` and `rpc` modules are not providing core building blocks, and do not belong to the `dpapi-core` crate.

crates/dpapi-core/src/str.rs
```diff
@@ -26,7 +29,15 @@ pub fn from_utf16_le(data: &[u8]) -> Result<String> {
 /// *Note*: this function automatically appends a NULL-char.
 pub fn encode_utf16_le(data: &str) -> Vec<u8> {
     data.encode_utf16()
-        .chain(std::iter::once(0))
+        .chain(core::iter::once(0))
         .flat_map(|v| v.to_le_bytes())
         .collect::<Vec<_>>()
 }
```
CBenoit, line 35: thought: Now that you have `str_utf16_len`, you may want to encode the utf16 string in-place using a `WriteCursor` instead.

TheBestTvarynka: done

crates/dpapi-core/src/decode.rs
```diff
@@ -0,0 +1,61 @@
...
+        .map_err(|err| DecodeError::invalid_field("", "uuid", "invalid data").with_source(err))
+}
+
+/// PDU that can be decoded from a binary input and provided context.
+pub trait DecodeWithContextOwned: Sized + NeedsContext {
+    /// Decodes PDU from a [`ReadCursor`] with provided context.
+    fn decode_with_context_owned(src: &mut ReadCursor<'_>, ctx: Self::Context<'_>) -> DecodeResult<Self>;
+}
+
+/// Fixed size of the srtructure.
```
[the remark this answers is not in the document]
TheBestTvarynka: done

crates/dpapi-core/src/padding.rs
TheBestTvarynka: wow, thank for the comprehensive explanation!
done

crates/dpapi-core/src/lib.rs
TheBestTvarynka: I thought about it. I explored the IronRDP crates structure more and decided to move structures encoding and decoding implementations into a separate crate: `dpapi-pdu`. Moreover, I added an `alloc` feature to the `dpapi-core` crate and put the `EncodeVec` extension trait behind it.

I just pushed a new commit. Can you have a look?
Surprisingly, it was easy to refactor. I think it's a sign of a good architecture 🙃

crates/dpapi-core/src/str.rs
```diff
@@ -27,11 +29,11 @@ pub fn from_utf16_le(data: &[u8]) -> Result<String> {
 /// Encodes str into a UTF-16 encoded byte array.
 ///
 /// *Note*: this function automatically appends a NULL-char.
-pub fn encode_utf16_le(data: &str) -> Vec<u8> {
+/// *Panic*: panics when cursor's internal buffer doesn't have enough space.
```
CBenoit, line 33: note: Typically this kind of information is specified in a dedicated section. See: https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure

CBenoit (approved):
LGTM! Excellent work! :100:

crates/dpapi-core/src/lib.rs
CBenoit: Thank you!!
It definitely is :wink:

crates/dpapi-core/Cargo.toml
CBenoit: Note that I still recommend against using `num-derive` and `num-traits` crates even in `dpapi-pdu`. But if you are sure it covers your needs properly, and if it’s really saving you trouble, it’s fine!

PR #410, opened 2025-04-03
feat(dpapi): WS tunneling support

TheBestTvarynka:
Hi,
I implemented WS tunneling support in this PR.

I tested the DPAPI on the following scenarios:

* Direct TCP transport.
* Tunneling over WS connection.
* Tunneling over WSS connection.

Everything works well.


crates/dpapi-ws/src/gateway_client.rs
TheBestTvarynka, line 1: You most likely will have questions about this module, so let me explain myself 🙂 

We don't know the target server port (proxy destination address) in advance. We obtain the RPC server port during EPT MAP exchange. So, we need obtain a session token for the obtained server port.
The `GatewayClient` in this module acquires the session token for the needed destination url.

We can reduce the code in this module if we provide the client with a web token, instead of user+password. This way, we'd need to do only HTTP requests for session tokens.

I find the way with provided web token more secure. What do you think?

ffi/src/dpapi/mod.rs
```diff
@@ -112,18 +115,19 @@ pub unsafe extern "system" fn DpapiProtectSecret(
...
+        //         sid,
+        //         root_key,
+        //         server,
+        //         username,
+        //         password.into(),
+        //         computer_name
+        //     ),
+        //     NTE_INTERNAL_ERROR
+        // );
+        let blob_data: Vec<u8> = todo!();
```
TheBestTvarynka, line 130: Now, DPAPI functions are async. But we need to call them synchronously here. We have two options:

* `tokio::spawn_blocking`
* use `maybe_async` crate for `dpapi` crate and generate sync functions.

I didn't try `maybe_async`, but I'm unsure if it will not cause problems with `Transport`.

What do you think?

TheBestTvarynka:
@CBenoit When will you have a chance to review it? 🙂

CBenoit:
@TheBestTvarynka I’ll review tomorrow! :slightly_smiling_face: 
The PR is quite big, and I think I’ll have to think over it a little bit!

crates/dpapi-dg/README.md
```diff
@@ -0,0 +1,3 @@
+# DPAPI Devolutions-Gateway
+
+The main purpose of this crate is to obtain the session token from Devolutions-Gateway using standartalone route.
```
CBenoit, line 3: issue: Already discussed on Slack, but the webapp routes are dedicated to the Devolutions Gateway standalone, and should not be relied upon.

suggestion: If this is just for testing purpose, you may consider support for [tokengen server](https://github.com/Devolutions/devolutions-gateway/tree/master/tools/tokengen). Note that this is a dev tool, and this is not something that we will ever use in production.

question: My understanding is that `dpapi-cli-client` is also a dev tool (testing)?

crates/dpapi-native-transport/Cargo.toml
TheBestTvarynka, line 1: we need this code in the `dpapi-cli-client` and in the `ffi` crates. So, I moved it into a separate crate

crates/dpapi-transport/src/connect_options.rs
```diff
@@ -0,0 +1,125 @@
...
+use uuid::Uuid;
+
+/// Type that represents a function for obtaining the sesion token.
+///
+/// We need it because we don't know the destination address in advance.
+///
+/// Parameters:
+/// * `Uuid` is the session id.
+/// * `Url` is the destination of the proxied connection.
+pub type GetSessionTokenFn = dyn Fn(Uuid, Url) -> Pin<Box<dyn Future<Output = std::io::Result<String>>>>;
```
TheBestTvarynka, line 14: As you can notice, I decided to use dynamic dispatch here. There are a few reasons behind this decision:

1. API simplicity.
2. Usage convenience. If I had chosen generics over trait object, then the user would have to specify `GetSessionTokenFn` type even when they do not plan to tunnel the connection.

crates/dpapi-dg/README.md
TheBestTvarynka: done

TheBestTvarynka:
@CBenoit I finished refactoring the code. Feel free to review whenever you want

crates/dpapi-transport/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+pub trait Transport {
+    /// A type that represents communication stream.
+    type Stream: LocalStream;
+
+    /// Connects to the target server.
+    fn connect(connect_options: &ConnectOptions) -> impl Future<Output = Result<Self::Stream>>;
+}
+
+/// Stream for reading and writing bytes.
+pub trait LocalStream {
```
CBenoit, line 23: question: What is `Local` for?

crates/dpapi/src/blob.rs
```diff
@@ -173,18 +170,18 @@ impl DpapiBlob {
                     unprotected_attrs: Optional::from(None),
                 })?)),
             },
-            &mut writer,
+            &mut *dst,
```
CBenoit, line 173: suggestion: It looks like this could be this:
```suggestion
            dst,
```

crates/dpapi/src/client.rs
```diff
@@ -189,10 +190,10 @@ struct GetKeyArgs<'server> {
     negotiate_config: NegotiateConfig,
 }
 
-#[instrument(level = "trace", ret)]
-fn get_key(
+async fn get_key<T: Transport>(
```
CBenoit, line 193: note: In `sspi`, we used a different approach to avoid asyncifying the API.
We used generators: https://github.com/Devolutions/sspi-rs/blob/3b02a72c5116445df9aa72cfc1d66072dfa71ef8/src/generator.rs

We avoid the need for a fully-fledged async runtime such as Tokio in the native DLL.
I think ideally we may want something like that, but it’s some work to move away from the current asyncified API…

crates/dpapi-transport/src/lib.rs
TheBestTvarynka: 🤔 I don't remember why I named it `Local`. I think I can rename it to just `Stream`

crates/dpapi/src/client.rs
CBenoit: Another note: I was working on a library for generators, but I think I was still not satisfied with something. Probably still an improvement over the one we vendored into sspi-rs: https://github.com/CBenoit/genoise

crates/dpapi-transport/src/lib.rs
TheBestTvarynka: > I think I can rename it to just `Stream`

done

crates/dpapi/src/client.rs
TheBestTvarynka: > I think ideally we may want something like that, but it’s some work to move away from the current asyncified API…

I figured out the generator's implementation. I can implement it. But I'm not sure if I should do it now. I think I need to ask Mark-André about this task urgency.

> Probably still an improvement over the one we vendored into sspi-rs

Interesting library. I read its code, and I agree, and it looks more cleaner then the `geenerator` module in `sspi-rs`

CBenoit: Discussed on Slack, decided to not refactor for now :heavy_check_mark:

tools/dpapi-cli-client/src/session_token.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+#[derive(Deserialize)]
+struct TokenResponse {
+    token: String,
+}
+
+/// Obtains the session token from the [tokengen server](https://github.com/Devolutions/devolutions-gateway/tree/master/tools/tokengen).
+///
+/// Paramers:
```
CBenoit, line 27: typo
```suggestion
/// Parameters:
```

tools/dpapi-cli-client/README.md
```diff
@@ -0,0 +1,5 @@
+# DPAPI cli client
+
+This is a simple DPAPI client for using the DPAPI. It can encrypt secrets or decrupt them using DPAPI. Run `dpapi-cli-client --help` for more information and usage manual.
```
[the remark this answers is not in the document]
TheBestTvarynka: I'm wondering why VS code didn't highlight it 🤔

src/rustls.rs
```diff
@@ -1,7 +1,8 @@
 #![cfg(feature = "__rustls-used")]
 
 /// Call this before using rustls.
-pub(crate) fn install_default_crypto_provider_if_necessary() -> Result<(), ()> {
+#[allow(clippy::result_unit_err)]
+pub fn install_default_crypto_provider_if_necessary() -> Result<(), ()> {
```
CBenoit, line 5: issue: This is not intended to be part of the public API of `sspi`.

suggestion: Not ideal, but can you add `#[doc(hidden)]`? Let’s at least hide it.

src/lib.rs
```diff
@@ -72,7 +72,7 @@ mod crypto;
 mod dns;
 mod kdc;
 mod krb;
-mod rustls;
+pub mod rustls;
```
CBenoit, line 75: issue: Let’s keep this module private. Following previous advice, re-export from lib.rs and ensure the item is properly hidden from the documentation.

ffi/src/winscard/scard_handle.rs
```diff
@@ -537,7 +537,7 @@ pub unsafe fn copy_io_request_to_scard_io_request(
 
     let pci_info_len = io_request.pci_info.len();
     // SAFETY: it's safe to deref because we've checked for null value above.
-    let scard_pci_info_len = unsafe { (*scard_io_request).cb_pci_length }.try_into()?;
+    let scard_pci_info_len = usize::try_from(unsafe { (*scard_io_request).cb_pci_length })?;
```
CBenoit, line 540: style: I recommend splitting on two lines instead of having the unsafe block inlined.

tools/dpapi-cli-client/src/session_token.rs
TheBestTvarynka: done

tools/dpapi-cli-client/README.md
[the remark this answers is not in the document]
TheBestTvarynka: > I'm wondering why VS code didn't highlight it 🤔

the spell checker extension was turned off...I don't know why. Anyway, I turned it on again and fixed many other typos. See this commit: https://github.com/Devolutions/sspi-rs/pull/410/commits/b0856bece2403dcd03d1f6f030c45ac1e670705e

src/rustls.rs
TheBestTvarynka: done

src/lib.rs
TheBestTvarynka: done

ffi/src/winscard/scard_handle.rs
TheBestTvarynka: done

ffi/src/dpapi/session_token.rs
```diff
@@ -0,0 +1,61 @@
...
+            let destination = CString::new(destination.as_str())
+                .map_err(|err| Error::new(ErrorKind::InvalidData, format!("invalid destination url: {:?}", err)))?;
+            let mut token_len = 2048;
+            let mut token_buf = vec![0; 2048];
+
+            // SAFETY: all function input parameters are valid because:
+            // * session_id is an object on stack.
+            // * destination is created (and validated) using `CString`.
+            // * token_buf is a non-empty Vec.
+            // * token len is a local variale.
```
CBenoit, line 34:
```suggestion
            // * token len is a local variable.
```

TheBestTvarynka: done

ffi/src/dpapi/session_token.rs
```diff
@@ -0,0 +1,61 @@
...
+                data3,
+                data4: *data4,
+            };
+
+            let destination = CString::new(destination.as_str())
+                .map_err(|err| Error::new(ErrorKind::InvalidData, format!("invalid destination url: {:?}", err)))?;
+            let mut token_len = 2048;
+            let mut token_buf = vec![0; 2048];
+
+            // SAFETY: all function input parameters are valid because:
```
CBenoit, line 30:
```suggestion
            // SAFETY:
            // As per safety preconditions, the C function pointer is safe to be called with valid parameters.
            // Parameters are valid:
```

ffi/src/dpapi/session_token.rs
```diff
@@ -0,0 +1,61 @@
...
+
+use dpapi_transport::GetSessionTokenFn;
+use ffi_types::Uuid as CUuid;
+use url::Url;
+use uuid::Uuid;
+
+use super::GetSessionTokenFn as CGetSessiontokenFn;
+
+/// This function wraps a C-function into a Rust closure which we can pass into the Rust API.
+pub fn session_token_fn(get_session_token: CGetSessiontokenFn) -> Box<GetSessionTokenFn> {
```
CBenoit:
```suggestion
/// This function wraps a C-function into a Rust closure which we can pass into the Rust API.
///
/// # Safety
///
/// The C function pointer must be safe to call provided parameters are valid.
pub unsafe fn session_token_fn(get_session_token: CGetSessiontokenFn) -> Box<GetSessionTokenFn> {
```

ffi/src/dpapi/session_token.rs
TheBestTvarynka: done

crates/dpapi-native-transport/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "dpapi-native-transport"
+version = "0.1.0"
+edition = "2024"
```
CBenoit, line 4: suggestion: Can you add the required meta data needed for publishing crates?

TheBestTvarynka: done (sorry for the delay, I had a meeting)

crates/dpapi-transport/Cargo.toml
```diff
@@ -1,7 +1,13 @@
 [package]
 name = "dpapi-transport"
 version = "0.1.0"
-edition = "2024"
+edition = "2021"
```
CBenoit, line 4: suggestion: Feel free to keep edition 2024 for new code. We’ll want to update to 2024 eventually.

crates/dpapi-native-transport/Cargo.toml
CBenoit: No problem! Thank you!

crates/dpapi-transport/Cargo.toml
TheBestTvarynka: I just wanted to be consistent. I know that crates can have different edition versions, but my inner voice says that I want it to be the same for all crates in the workspace 😄

PR #412, opened 2025-04-07
test(fuzz): add fuzzing for `dpapi-pdu` structures

TheBestTvarynka:
Hi,
I decided to add fuzzing to `dpapi-pdu` crate. Because we have a lot of decoding/encoding there.

It wasn't hard to implement it. It was much harder to fix bugs found by fuzzing.

PR #415, opened 2025-04-14
test(fuzz): add fuzzing for `dpapi-pdu` structures

TheBestTvarynka:
Hi,
This is the same PR as #412, but with addressed comments. I reverted these changes in #410 and submitting them again.

This PR should be merged after #410.


CBenoit:
LGTM!

crates/dpapi-fuzzing/src/oracles/round_trip.rs
```diff
@@ -0,0 +1,13 @@
+use dpapi_core::{ReadCursor, WriteBuf};
+
+use crate::generator::AnyStruct;
+
+pub fn round_trip(any: AnyStruct) {
+    let mut buf = WriteBuf::new();
+
+    if let Ok(name) = any.encode(&mut buf) {
```
CBenoit, line 8: question: By the way, we have structs that can't be encoded without it being a bug?

PR #419, opened 2025-04-22
feat(dpapi): add web client; 

TheBestTvarynka:
Hi,
I implemented a DPAPI web client in this PR.

I tested the implementation on a small `.js` script which calls the WASM DPAPI module functions.


crates/dpapi-pdu/src/rpc/pdu.rs
```diff
@@ -573,12 +573,12 @@ impl DecodeWithContextOwned for PduData {
...
-                "invalid PDU data",
-                "PDU header or PDU data is invalid: too much data to decode",
-            ));
-        }
+        // if !buf.is_empty() {
+        //     return Err(DecodeError::other(
+        //         "invalid PDU data",
+        //         "PDU header or PDU data is invalid: too much data to decode",
+        //     ));
+        // }
```
CBenoit, line 581: Leftover?

TheBestTvarynka: Yes, +
This PR still has a few places that need to be cleaned up 🙏 (this is why it is draft 😅)

TheBestTvarynka:
Small testing report

I checked the implementation on the following configurations:

1. CLI client: direct TCP connection.
2. CLI client: WS tunneled connection + KDC proxy.
3. WEB client: WS tunneled connection + KDC proxy. I wrote a simple HTML file (I can attach it if you need) with embedded script which calls our DPAPI WASM module and run it in the browser. It works: 
image

Everything works well!

> I wrote a simple HTML file with embedded script which calls our DPAPI WASM module and run it in browser.

Before that, I was testing it using only a script file and node.js runtime. I found out that it constantly throws an exception when running in node.js runtime:

image

I investigated it for some time. As I understood, Rust drops the WebSocket object before it gets a change to handle the `Socket.onSocketClose`. It is happening only inside node.js runtime. The code works well inside the browser environment.

I don't think it's a problem for Devolutions, but it's better to be aware of it.

crates/dpapi-native-transport/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "dpapi-native-transport"
 version = "0.1.0"
-edition = "2024"
+edition = "2021"
```
CBenoit, line 4: note: I think it’s fine to stick to 2024. We’ll migrate to 2024 eventually, so we may as well reduce the future effort for it.

PR #430, opened 2025-05-15
refactor(ffi): sspi: enable lints for unsafe code

TheBestTvarynka:
Hi,
I refactored FFI code in this PR:

* Enabled `unsafe_op_in_unsafe_fn` and `clippy::undocumented_unsafe_blocks` lints for all modules in the FFI crate.
* Fixed warnings and compilation errors.

Do not be scared of big number of lines changed. They are mostly `SAFETY` comments 🙂

### DIFF
--- a/ffi/src/dpapi/mod.rs
+++ b/ffi/src/dpapi/mod.rs
@@ -182,7 +182,7 @@ pub unsafe extern "system" fn DpapiProtectSecret(
             return NTE_INTERNAL_ERROR;
         }
 
-        // SAFETY: Memory allocation should be safe. Moreover, we check for the null value below.
+        // SAFETY: Memory allocation is safe Moreover, we check for the null value below.
         let blob_buf = unsafe { libc::malloc(blob_data.len()) as *mut u8 };
         if blob_buf.is_null() {
             error!("Failed to allocate memory for the output DPAPI blob: blob buf pointer is NULL");
@@ -327,7 +327,7 @@ pub unsafe extern "system" fn DpapiUnprotectSecret(
             return NTE_INTERNAL_ERROR;
         }
 
-        // SAFETY: Memory allocation should be safe. Moreover, we check for the null value below.
+        // SAFETY: Memory allocation is safe Moreover, we check for the null value below.
         let secret_buf = unsafe { libc::malloc(secret_data.as_ref().len()) as *mut u8 };
         if secret_buf.is_null() {
             error!("Failed to allocate memory for the output DPAPI blob: blob buf pointer is NULL.");
--- a/ffi/src/dpapi/mod.rs
+++ b/ffi/src/dpapi/mod.rs
@@ -182,7 +182,7 @@ pub unsafe extern "system" fn DpapiProtectSecret(
             return NTE_INTERNAL_ERROR;
         }
 
-        // SAFETY: Memory allocation is safe Moreover, we check for the null value below.
+        // SAFETY: Memory allocation is safe. Moreover, we check for the null value below.
         let blob_buf = unsafe { libc::malloc(blob_data.len()) as *mut u8 };
         if blob_buf.is_null() {
             error!("Failed to allocate memory for the output DPAPI blob: blob buf pointer is NULL");
@@ -327,7 +327,7 @@ pub unsafe extern "system" fn DpapiUnprotectSecret(
             return NTE_INTERNAL_ERROR;
         }
 
-        // SAFETY: Memory allocation is safe Moreover, we check for the null value below.
+        // SAFETY: Memory allocation is safe. Moreover, we check for the null value below.
         let secret_buf = unsafe { libc::malloc(secret_data.as_ref().len()) as *mut u8 };
         if secret_buf.is_null() {
             error!("Failed to allocate memory for the output DPAPI blob: blob buf pointer is NULL.");
--- a/ffi/src/sspi/common.rs
+++ b/ffi/src/sspi/common.rs
@@ -265,7 +265,6 @@ pub type VerifySignatureFn = extern "system" fn(PCtxtHandle, PSecBufferDesc, u32
 #[no_mangle]
 pub unsafe extern "system" fn FreeContextBuffer(pv_context_buffer: *mut c_void) -> SecurityStatus {
     // NOTE: see https://github.com/Devolutions/sspi-rs/pull/141 for rationale behind libc usage.
-    check_null!(pv_context_buffer);
     // SAFETY: Memory deallocation should be safe.
     unsafe {
         libc::free(pv_context_buffer);
--- a/ffi/src/sspi/common.rs
+++ b/ffi/src/sspi/common.rs
@@ -29,7 +29,10 @@ pub unsafe extern "system" fn FreeCredentialsHandle(ph_credential: PCredHandle) 
     let cred_handle = unsafe { (*ph_credential).dw_lower as *mut CredentialsHandle };
     check_null!(cred_handle);
 
-    // SAFETY: `cred_handle` is not null. We've checked for this above. So it's safe to call the function.
+    // SAFETY: `cred_handle` is not null. We've checked for this above.
+    // We create and allocate credentials handles using `Box::into_raw`. Thus,
+    // it is safe to deallocate them using `Box::from_raw`.
+    // The user have to ensure that the credentials handle was created by us.
     let _cred_handle = unsafe { Box::from_raw(cred_handle) };
 
     0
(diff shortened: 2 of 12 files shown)


src/cert_utils.rs
TheBestTvarynka, line 1: I refactored this module because the FFI layer uses some of its functions, but there is still a place for improvement. I think it's better to do them later in a separate PR.

CBenoit (commented):
[In music](https://www.youtube.com/watch?v=G7LJC9vJluU) :musical_note: (okay, there is no C here :stuck_out_tongue: )

More seriously though, very happy that you went through that!! :pray: 
Our unsafe code documentation here really needed some love.

I’ll poke @thenextman, @allan2, @RRRadicalEdward, @pacmancoder and @irvingoujAtDevolution, who may also write unsafe code soon, so that everyone is on the same page.

I have a few general guidelines on how I would like the unsafe code to be written and documented in the future:

- This is a Rust community convention: unsafe blocks should be documented using `SAFETY:` comments, and each unsafe block should only have a single unsafe operation.

- Explain why the code is "sound", detailing the invariants and preconditions that are maintained.

- Avoid assumptions without justifications.
	- Do not use phrases like "we assume this is safe because…" or "this should be safe".
	- Provide concrete evidence or reasoning that justifies the safety of the operation.

- Explicitly reference type invariants and prior checks when safety relies on it.

	```rust
	// SAFETY: index is guaranteed to be within bounds due to the prior check.
	unsafe { array.get_unchecked(index) };
	```

- Address each safety requirement individually, using a bullet list.

	```rust
	// SAFETY:
	// - ptr is non-null and properly aligned.
	// - The memory region ptr points to is valid for reads of len elements.
	unsafe { std::slice::from_raw_parts(ptr, len) };
	```

- Document FFI boundaries clearly, but if the only safety requirements are standard (e.g.: non-dangling pointers), for consistency, use the following concise comment verbatim:

    ```rust
	// SAFETY: FFI call with no outstanding preconditions.
	unsafe { straightforward_ffi_function(ptr) };
    ```

- You don’t have to document how the resulting value is safely used later.
	- Really focus only on why the current unsafe operation is sound.
	- Subsequent operations should be documented as appropriate in the following code.

- It’s welcomed to document how and why the function is used, but this should not be part of the "safety" comment.
    - Write a separate paragraph out of the safety comment, typically above the safety comment.

    ```rust
    // Passing None to GetModuleHandleW requests the handle to the
    // current process main executable module
    //
    // SAFETY: FFI call with no outstanding preconditions.
    let h_module: HMODULE = unsafe { GetModuleHandleW(None)? };
    ```

- To keep the comment concise, do not use phrases like "this is safe because…". Go straight to the point (e.g.: "pointer is guaranteed not null due to the prior check").


- For readability and clarity, keep the unsafe block as small as possible, and untangled from other safe code. Feel free to use intermediate variables.

    ```rust
    // SAFETY: …
    let intermediate_variable = unsafe { transform_credentials_handle(credentials_handle) };
    
    match intermediate_variable { … }
    ```

- If a precondition must be upheld by the caller, mark the function as `unsafe` and document it with a `# Safety` section listing the invariants.

    ```rust
    /// Returns the length in characters of the C-string.
    ///
    /// # Safety
    ///
    /// - `s` must be a valid, null-terminated C-string.
    unsafe fn strlen(s: *const u8) -> usize { … }
    ```

    (Of course, using a bullet list is not necessary when there is only a single precondition.)

ffi/src/dpapi/mod.rs
```diff
@@ -182,7 +182,7 @@ pub unsafe extern "system" fn DpapiProtectSecret(
             return NTE_INTERNAL_ERROR;
         }
 
-        // SAFETY: Memory allocation should be safe. Moreover, we check for the null value below.
+        // SAFETY: Memory allocation is safe. Moreover, we check for the null value below.
```
CBenoit, line 185: > You don’t have to document how the resulting value is safely used later.

ffi/src/sspi/common.rs
```diff
@@ -25,10 +25,15 @@ use crate::utils::into_raw_ptr;
...
+    // SAFETY: `ph_credentials` is not null. We've checked for this above.
+    let cred_handle = unsafe { (*ph_credential).dw_lower as *mut CredentialsHandle };
     check_null!(cred_handle);
 
-    let _cred_handle = Box::from_raw(cred_handle);
+    // SAFETY: `cred_handle` is not null. We've checked for this above.
+    // We create and allocate credentials handles using `Box::into_raw`. Thus,
+    // it is safe to deallocate them using `Box::from_raw`.
+    // The user have to ensure that the credentials handle was created by us.
+    let _cred_handle = unsafe { Box::from_raw(cred_handle) };
```
CBenoit, line 36: > Address each safety requirement individually, using a bullet list.

Also, "the user have to ensure that the credentials handle was created by us." is more a `# Safety` section that should be in the documentation of the function, as it express a precondition of the `FreeCredentialsHandle` function. In turn, this precondition should be used to justify that the handle was indeed allocated using `Box`.

> If a precondition must be upheld by the caller, mark the function as unsafe and document it with a # Safety section listing the invariants.

ffi/src/sspi/common.rs
```diff
@@ -123,16 +149,27 @@ pub unsafe extern "system" fn CompleteAuthToken(
...
+        // * the value behind `ph_context` must be initialized by ourself: the user does not have to create the [CtxHandle] values ​​themselves.
+        // * other parameters are type checked.
+        let mut sspi_context_ptr = try_execute!(unsafe { p_ctxt_handle_to_sspi_context(
             &mut ph_context,
             None,
             &CredentialsAttributes::default()
-        ))
-        .as_mut()
-        .expect("security context pointer cannot be null");
+        )});
```
CBenoit, line 163: > For readability and clarity, keep the unsafe block as small as possible, and untangled from other safe code. Feel free to use intermediate variables.

ffi/src/sspi/common.rs
```diff
@@ -123,16 +149,27 @@ pub unsafe extern "system" fn CompleteAuthToken(
...
-        let raw_buffers = from_raw_parts((*p_token).p_buffers, (*p_token).c_buffers as usize);
-        let mut buffers = p_sec_buffers_to_security_buffers(raw_buffers);
+        // SAFETY: It's safe to call the `as_mut` function, because `sspi_context_ptr` is a local pointer,
+        // which is initialized by the `p_ctx_handle_to_sspi_context` function. Thus, the value behind this pointer is valid.
+        let sspi_context = unsafe { sspi_context_ptr.as_mut() };
+
+        // SAFETY: This function is safe to call because `p_buffers` is not null. We've checked this above.
+        let raw_buffers = unsafe { from_raw_parts((*p_token).p_buffers, (*p_token).c_buffers as usize) };
+        // SAFETY: This function is safe to call because `raw_buffers` is type checked. All other guarantees must be provided by user.
+        let mut buffers = unsafe { p_sec_buffers_to_security_buffers(raw_buffers) };
```
CBenoit, line 172: > To keep the comment concise, do not use phrases like "this is safe because…". Go straight to the point (e.g.: "pointer is guaranteed not null due to the prior check").

CBenoit (approved):
Here is some feedback on this PR, based on my comment above. I didn’t comment on every occurrence, but I thought it would be a good case study.

Of course, this PR is a clear, net improvement over the status quo, and I see no reason to not merge it now! Feel free to send follow up PRs as necessary, or simply keep in mind the guidelines I enumerated for future reference :slightly_smiling_face: 

Thank you!

TheBestTvarynka:
@CBenoit Thank you for such extended answer! 💟 I appreciate time and effort you put in guidelines 😊 I will improve safety comments when I have time

During working on this PR, I found one more issue besides using uninitialized data:

```rust
context.dw_lower = into_raw_ptr(sspi_context) as c_ulonglong;
```

When we cast pointer to integer we lose pointer provenance. I read [`std::ptr`](https://doc.rust-lang.org/std/ptr/index.html) module documentation carefully. I suppose we have two ways how to fix it:

- use [`expose_provenance`](https://doc.rust-lang.org/std/primitive.pointer.html#method.expose_provenance) and [with_exposed_provenance](https://doc.rust-lang.org/std/ptr/fn.with_exposed_provenance.html).
- use a global `HashMap` where the key is a number and the value is a reference to the handle reference (or even the handle itself).

Personally, I prefer the second approach. However, it is open to discussion and should be fixed in the future.

TheBestTvarynka:
> [In music](https://www.youtube.com/watch?v=G7LJC9vJluU) 🎵 (okay, there is no C here 😛 )

I liked this musical break 😆

PR #435, opened 2025-05-26
refactor(sspi): Kerberos client implementation

TheBestTvarynka:
Hi,
I refactored the Kerberos client implementation and added new test to check the Kerberos U2U negotiation.

### DIFF
--- a/src/kerberos/client/mod.rs
+++ b/src/kerberos/client/mod.rs
@@ -190,7 +190,7 @@ pub async fn initialize_security_context<'a>(
 
             let as_rep = as_exchange(client, yield_point, &kdc_req_body, pa_data_options).await?;
 
-            info!("AS exchange finished successfully.");
+            debug!("AS exchange finished successfully.");
 
             client.realm = Some(as_rep.0.crealm.0.to_string());
 
@@ -261,7 +261,7 @@ pub async fn initialize_security_context<'a>(
             let tgs_rep: KrbResult<TgsRep> = KrbResult::deserialize(&mut d)?;
             let tgs_rep = tgs_rep?;
 
-            info!("TGS exchange finished successfully");
+            debug!("TGS exchange finished successfully");
 
             let session_key_2 = extract_session_key_from_tgs_rep(&tgs_rep, &session_key_1, &client.encryption_params)?;
 
@@ -293,7 +293,7 @@ pub async fn initialize_security_context<'a>(
                 warn!("Kerberos ApReq Authenticator checksum GSS_C_DELEG_FLAG is not supported. Turning it off...");
                 flags.remove(GssFlags::GSS_C_DELEG_FLAG);
             }
-            info!(?flags, "ApReq Authenticator checksum flags");
+            debug!(?flags, "ApReq Authenticator checksum flags");
 
             let mut checksum_value = ChecksumValues::default();
             checksum_value.set_flags(flags);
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -44,11 +44,13 @@ sha1 = { version = "0.10", default-features = false }
...
-picky-asn1 = "0.10"
-picky-asn1-der = "0.5"
-picky-asn1-x509 = "0.14"
-picky-krb = "0.9"
+
+picky = { path = "../picky-rs/picky", default-features = false }
+picky-asn1 = { path = "../picky-rs/picky-asn1" }
+picky-asn1-der = { path = "../picky-rs/picky-asn1-der" }
+picky-asn1-x509 = { path = "../picky-rs/picky-asn1-x509" }
+picky-krb = { path = "../picky-rs/picky-krb" }
```
TheBestTvarynka, line 52: https://github.com/Devolutions/picky-rs/pull/369

src/kerberos/client/generators.rs
```diff
@@ -68,9 +68,9 @@ const MD5_CHECKSUM_TYPE: [u8; 1] = [0x07];
 // https://www.rfc-editor.org/rfc/rfc4120#section-5.4.1
 pub const DEFAULT_AS_REQ_OPTIONS: [u8; 4] = [0x00, 0x81, 0x00, 0x10];
 
-// Renewable, Canonicalize, Enc-tkt-in-skey are on by default
+// Renewable, Canonicalize.
 // https://www.rfc-editor.org/rfc/rfc4120#section-5.4.1
-const DEFAULT_TGS_REQ_OPTIONS: [u8; 4] = [0x00, 0x81, 0x00, 0x08];
+const DEFAULT_TGS_REQ_OPTIONS: [u8; 4] = [0x00, 0x81, 0x00, 0x00];
```
TheBestTvarynka, line 73: I found a mistake here after writing new tests. We don't need to set the `enc-tkt-in-skey` flag by default. It should be set only when the Kerberos U2U is negotiated

Cargo.toml
CBenoit: Published!

TheBestTvarynka: @CBenoit I don't see a new version of the `picky-krb` on crates.io: https://crates.io/crates/picky-krb/versions

CBenoit: Oops… Looks like a bug, but it got moving again now

TheBestTvarynka:
@CBenoit the PR is ready for review :)

src/kerberos/client/mod.rs
```diff
@@ -1,2 +1,435 @@
...
+                        }),
+                        with_pre_auth: false,
+                        authenticator_nonce: OsRng.gen::<[u8; 4]>(),
+                    }))
+                }
+            };
+
+            let as_rep = as_exchange(client, yield_point, &kdc_req_body, pa_data_options).await?;
+
+            info!("AS exchange finished successfully.");
```
CBenoit, line 193: issue: Please avoid using INFO-level logs in libraries, as it tends to pollute usual logs of applications such as IronRDP and Devolutions Gateway. Let’s use DEBUG and TRACE only in the libraries. Feel free to use INFO in the binaries such as the `ffi` crate and other examples.

src/kerberos/server/mod.rs
CBenoit, line 1: nitpick: Appears to be an empty file now

PR #436, opened 2025-05-27
fix: lower info-level logs to debug-level

CBenoit:
Libraries should avoid issuing info-level logs.
We should stick to debug and trace levels.

cc @TheBestTvarynka


TheBestTvarynka (approved):
LGTM!

PR #441, opened 2025-06-05
feat(sspi): migrate from `winreg` to `windows-registry` crate

### LINKED ISSUE #347: Migrate from `winreg` to `windows-registry`
Let’s align with the broader `windows` ecosystem that we are already using. `windows-registry` is the “official” crate that stays up to date with the Microsoft-maintained Windows metadata. Error handling will be more consistent across all windows crates.

TheBestTvarynka:
I had some time between reviews. I scrolled through issues and decided to quickly fix #347.

closes #347

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,7 +29,7 @@ updates:
       windows:
         patterns:
           - "windows*"
-          - "windows-registry"
+          - "winreg"
           - "winapi"
       patch:
         dependency-type: "production"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -29,7 +29,7 @@ updates:
       windows:
         patterns:
           - "windows*"
-          - "winreg"
+          - "windows-registry"
           - "winapi"
       patch:
         dependency-type: "production"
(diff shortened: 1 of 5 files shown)


Cargo.lock
```diff
@@ -1443,7 +1443,7 @@ dependencies = [
  "socket2",
  "widestring",
  "windows-sys 0.48.0",
- "winreg 0.50.0",
+ "winreg",
```
TheBestTvarynka, line 1446: Unfortunatelly, we still have `winreg` in our dependency tree because of the `ipconfig` crate. it hasn't updated in two years (https://github.com/liranringel/ipconfig) and used by the DNS resolver we use (`hickory-resolver`)

.github/dependabot.yml
```diff
@@ -29,7 +29,7 @@ updates:
       windows:
         patterns:
           - "windows*"
-          - "winreg"
+          - "windows-registry"
```
CBenoit, line 32: It’s best to change that in a separate, dedicated PR, because it takes a special approval that I can’t give you directly. But in this case, I believe the only required change is removing `winreg`, because `windows*` is already covering `windows-registry`.

Cargo.lock
CBenoit: It is what it is…! Maybe one day this will be removed from the dependency tree. At least, we only have one winreg instead of two :joy:

CBenoit (approved):
LGTM! But open a separate PR for the dependabot configuration removing `winreg`

.github/dependabot.yml
TheBestTvarynka: I reverted this change. I'll open a separate PR soon.

Issue #433, opened 2025-05-24
Negotiate package tries Kerberos even when disabled and unsupported


AvivNaaman:
Hello,

I've been using the crate for authenticating a client against a remote server. I have started using the `Negotiate` SSP package, and I have noticed some issues with its implementations, which I'd like to resolve. Most of those issues are relevant for non-domain environments, and it might be a misconfiguration of the provider by me.

The initialization of the provided may be seen [here](https://github.com/AvivNaaman/smb-rs/blob/6a80bb65b941c664e0c9d2df52d2a94463737285/smb/src/session/authenticator.rs#L39) in my crate's source code.

### Kerberos is disabled in the options of the SSP
Providing a `package_list` argument to `NegotiateConfig` with the value `"ntlm,!kerberos,!pku2u"`, I am still seeing logs of the crate that indicate 3 attempts to perform a KDC detection using DNS(?):
```
[2025-05-24T14:57:07Z DEBUG smb::session] Setting up session for user LocalAdmin.
[smb/src/session/authenticator.rs:41:18] Self::get_available_ssp_pkgs(&conn_info.config.auth_methods) = "ntlm,!kerberos,!pku2u"
[2025-05-24T14:57:07Z INFO  sspi::negotiate] acquire_credentials_handle_impl; protocol="NTLM"
[2025-05-24T14:57:07Z INFO  sspi::negotiate] negotiate_protocol; username="LocalAdmin" domain="" protocol="NTLM"
[2025-05-24T14:57:07Z INFO  sspi::kdc] detect_kdc_hosts; domain=""
[2025-05-24T14:57:07Z DEBUG sspi::kdc] detect_kdc_hosts_from_system; domain=""
[2025-05-24T14:57:07Z DEBUG sspi::kdc] return=[]
[2025-05-24T14:57:07Z DEBUG sspi::dns] detect_kdc_hosts_from_dns; domain=""
[2025-05-24T14:57:08Z ERROR sspi::dns] Timeout when reading DNS query error=deadline has elapsed
[2025-05-24T14:57:09Z ERROR sspi::dns] Timeout when reading DNS query error=deadline has elapsed
[2025-05-24T14:57:09Z DEBUG sspi::dns] return=[]
[2025-05-24T14:57:09Z INFO  sspi::kdc] return=[]
[2025-05-24T14:57:09Z INFO  sspi::negotiate] return=Ok(())
[...]

PR #447, opened 2025-06-10
fix: Negotiate attempts KDC detection even when Kerberos is disabled

AvivNaaman:
See #433


CBenoit:
Published in `sspi v0.15.11`

AvivNaaman:
Thanks @CBenoit -- you're awesome

PR #457, opened 2025-06-24
feat(sspi): server-side Kerberos fixes

TheBestTvarynka:
Hi,
This PR contains a small patch for better support of server-side Kerberos. I wrote it during debugging and dev-testing the RDP proxy Devolutions-Gateway feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,11 +44,13 @@ sha1 = { version = "0.10", default-features = false }
 sha2 = "0.10"
 num-derive = "0.4"
 num-traits = { version = "0.2", default-features = false }
-picky = { version = "7.0.0-rc.12", default-features = false }
-picky-asn1 = "0.10"
-picky-asn1-der = "0.5"
-picky-asn1-x509 = "0.14"
-picky-krb = "0.10"
+
+picky = { path = "../picky-rs/picky", default-features = false }
+picky-asn1 = { path = "../picky-rs/picky-asn1" }
+picky-asn1-der = { path = "../picky-rs/picky-asn1-der" }
+picky-asn1-x509 = { path = "../picky-rs/picky-asn1-x509" }
+picky-krb = { path = "../picky-rs/picky-krb" }
+
 tokio = "1.44"
 ffi-types = { path = "crates/ffi-types" }
 winscard = { version = "0.2", path = "crates/winscard" }
--- a/src/credssp/mod.rs
+++ b/src/credssp/mod.rs
@@ -248,7 +248,7 @@ impl CredSspClient {
                 .expect("CredSsp client mode should never be empty")
             {
                 ClientMode::Negotiate(negotiate_config) => Some(CredSspContext::new(SspiContext::Negotiate(
-                    Negotiate::new(negotiate_config)?,
+                    Negotiate::new_client(negotiate_config)?,
                 ))),
                 ClientMode::Kerberos(kerberos_config) => Some(CredSspContext::new(SspiContext::Kerberos(
                     Kerberos::new_client_from_config(kerberos_config)?,
@@ -460,7 +460,7 @@ impl<C: CredentialsProxy<AuthenticationData = AuthIdentity> + Send> CredSspServe
                 .expect("CredSsp client mode should never be empty")
             {
                 ServerMode::Negotiate(neg_config) => Some(CredSspContext::new(SspiContext::Negotiate(
-                    try_cred_ssp_server!(Negotiate::new(neg_config), ts_request),
+                    try_cred_ssp_server!(Negotiate::new_server(neg_config), ts_request),
                 ))),
                 ServerMode::Kerberos(kerberos_mode) => {
                     let (kerberos_config, server_properties) = *kerberos_mode;
--- a/src/credssp/mod.rs
+++ b/src/credssp/mod.rs
@@ -60,10 +60,10 @@ macro_rules! try_cred_ssp_server {
                 let error = crate::Error::from(e);
                 $ts_request.error_code = Some(construct_error(&error));
 
-                return Err(ServerError {
+                return Err(Box::new(ServerError {
                     ts_request: $ts_request,
                     error,
-                });
+                }));
             }
         }
     };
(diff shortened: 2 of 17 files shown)


Cargo.toml
```diff
@@ -44,11 +44,13 @@ sha1 = { version = "0.10", default-features = false }
...
-picky-asn1 = "0.10"
-picky-asn1-der = "0.5"
-picky-asn1-x509 = "0.14"
-picky-krb = "0.10"
+
+picky = { path = "../picky-rs/picky", default-features = false }
+picky-asn1 = { path = "../picky-rs/picky-asn1" }
+picky-asn1-der = { path = "../picky-rs/picky-asn1-der" }
+picky-asn1-x509 = { path = "../picky-rs/picky-asn1-x509" }
+picky-krb = { path = "../picky-rs/picky-krb" }
```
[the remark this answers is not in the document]
TheBestTvarynka: Thanks!

src/credssp/mod.rs
```diff
@@ -437,13 +437,12 @@ impl<C: CredentialsProxy<AuthenticationData = AuthIdentity> + Send> CredSspServe
...
 
-    #[allow(clippy::result_large_err)]
     #[instrument(fields(state = ?self.state), skip_all)]
     pub fn process(
         &mut self,
         ts_request: TsRequest,
-    ) -> Generator<NetworkRequest, crate::Result<Vec<u8>>, Result<ServerState, ServerError>> {
-        Generator::<NetworkRequest, crate::Result<Vec<u8>>, Result<ServerState, ServerError>>::new(
+    ) -> Generator<NetworkRequest, crate::Result<Vec<u8>>, Result<ServerState, Box<ServerError>>> {
+        Generator::<NetworkRequest, crate::Result<Vec<u8>>, Result<ServerState, Box<ServerError>>>::new(
```
CBenoit, line 445: question: Why did you box the error? Is it because it makes the return type too big otherwise?

src/kerberos/client/extractors.rs
```diff
@@ -171,6 +171,7 @@ pub fn extract_status_code_from_krb_priv_response(
 }
 
 /// Extracts [ApRep] from the [NegTokenTarg1] .
+#[instrument(ret)]
```
CBenoit, line 174: issue: I don’t remember the syntax exactly, but can you verify this is not logging at INFO-level? Use either DEBUG or TRACE in the library.

tests/sspi/client_server/kerberos/kdc.rs
```diff
@@ -32,7 +33,7 @@ use rand::{Rng, RngCore};
 use sspi::kerberos::KERBEROS_VERSION;
 use time::{Duration, OffsetDateTime};
 
-pub const MAX_TIME_SKEW: Duration = Duration::minutes(3);
+pub const MAX_TIME_SKEW: StdDuration = StdDuration::from_secs(3);
```
CBenoit, line 36: suggestion: You may use either time::Duration or std::time::Duration, but there is no need to mix them both.

src/credssp/mod.rs
TheBestTvarynka: 🤔 I did not think about it. Good idea! done

tests/sspi/client_server/kerberos/kdc.rs
TheBestTvarynka: > but there is no need to mix them both.

I don't want to force users to have the `time` crate installed, so I decided to use `std::time::Duration` instead. It is used only in `ServerProperties::max_time_skew` (because it is a part of the public API). It is also convenient, because `time::Duration` implements addition/comparison with `std::time::Duration`

src/auth_identity.rs
```diff
@@ -170,6 +170,14 @@ impl AuthIdentityBuffers {
...
         self.user.is_empty()
     }
+
+    pub fn from_utf8(user: &str, domain: &str, password: &str) -> Self {
+        Self {
+            user: utils::string_to_utf16(user),
+            domain: utils::string_to_utf16(domain),
+            password: utils::string_to_utf16(password).into(),
+        }
+    }
```
CBenoit, line 180: thought: I think we should cleanup our UTF-16-related code using the widestring crate. It would be easier and less error-prone (better defined) than working with `Vec<u8>`.

tests/sspi/client_server/kerberos/kdc.rs
TheBestTvarynka: > I prefer `std::time::Duration` over renaming types, unless there is really no other better alternative.

done (I accidentally refactored it in my last commits 🙃)

src/auth_identity.rs
CBenoit: Makes sense

src/kerberos/config.rs
```diff
@@ -75,3 +77,25 @@ impl KerberosConfig {
...
+    fn new_instance(&self) -> Result<NegotiatedProtocol> {
+        Ok(NegotiatedProtocol::Kerberos(Kerberos::new_server_from_config(
+            self.kerberos_config.clone(),
+            self.server_properties.clone(),
+        )?))
+    }
+
+    fn box_clone(&self) -> Box<dyn ProtocolConfig> {
+        Box::new(self.clone())
+    }
```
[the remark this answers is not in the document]
CBenoit: I’m talking about this function itself. Since you already have the concrete type, you can:
- Clone and box yourself
- Clone the box

Maybe you added this function so you don’t have to cast with `as Box<dyn ProtocolConfig>`?

In that case, I would suggest a function `into_erased` such as:
```rust
fn into_erased(self) -> Box<dyn ProtocolConfig> {
    Box::new(self)
}
```
or
```rust
fn into_erased(self: Box<Self>) -> Box<dyn ProtocolConfig> {
    self
}
```

That is more explicit about the intention I think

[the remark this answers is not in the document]
CBenoit: In that case, I question whether we really need that in the public API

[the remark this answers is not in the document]
TheBestTvarynka: I just realized that I cannot make the `ProtocolConfig` trait private because the `NegotiateConfig::new` accepts `Box<dyn ProtocolConfig + Send>`. But I can seal this trait.

> my suggestion would be to use one of the approach I mentioned with `into_erased`

It allows to _convert_ the concrete type to `dyn ProtocolConfig`. But it does not solve the need of cloning the value in the `impl Clone for NegotiateConfig` implementation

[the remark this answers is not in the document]
TheBestTvarynka: But I do not see how `into_erased` will help us. The `NegotiateConfig` contains `protocol_config: Box<dyn ProtocolConfig>`, and we need to clone this value in the `Clone for NegotiateConfig` implementation. I do not know how `into_erased` will help us here because it moves the value into the method to return the boxed value...

PR #463, opened 2025-07-04
fix(sspi): Kerberos server WRAP token generation and validation

TheBestTvarynka:
Hi,
I fixed Kerberos server WRAP token generation and validation. I discovered these problems during dev-testing using `mstscex`.


TheBestTvarynka:
> By the way, would it make sense to add regression tests for this kind of stuff?

These changes also have improved the Kerberos client implementation: stricter checks during the acceptance of incoming messages. So, previously invalid messages will no longer be accepted, and our integration tests will catch such errors. This is why I added the `fake_server_properties` function: without it, existing tests fail.

PR #464, opened 2025-07-04
fix(sspi): Kerberos server MIC token generation and validation

TheBestTvarynka:
Hi,
I fixed Kerberos server MIC token generation and validation. I discovered these problems during dev-testing using `mstscex`.


CBenoit (approved):
Excellent job! Thank you!

PR #470, opened 2025-07-22
chore: add lints.workspace = true to every crate

RRRadicalEdward:


Cargo.toml
```diff
@@ -1,3 +1,5 @@
+lints.workspace = true
+
```
CBenoit, line 2: issue: It seems to be placed at a strange place

RRRadicalEdward: I remember that it's the only place in this file where it works. I was struggling with setting it, but I found a project that was setting it at the beginning of the file, and it worked in sspi-rs as well.

RRRadicalEdward: Done. Thank you

Issue #240, opened 2024-05-13
Explicitly adding channel bindings on the client


inejge:
In the current API I see no way to add channel bindings to an `Ntlm` context when authenticating a client. (Motivation: AD can be configured to refuse authentication form an LDAP client doing an NTLM bind over a TLS connection if the channel bindings AvP is not supplied.)

Adding a method to `Ntlm` which sets the `channel_bindings` field to the supplied `ChannelBindings` struct does the right thing protocol-wise, i.e., the channel bindings AvP is correctly calculated and added to the AUTHENTICATE message. Would a PR with this kind of addition be appropriate, or is there another preferred way to add this functionality to the library?

CBenoit:
Hi!

PRs are absolutely welcomed, and encouraged!!

About the suggested change, however, my guess is that it was not intended to override the `channel_bindings` field. It is already set internally in some code paths for NTML challenge and NTLM authenticate:

https://github.com/Devolutions/sspi-rs/blob/55ae932c70dd47b927096f9d6319b96ebe8aea26/src/ntlm/mod.rs#L275

If we add a method to set the `channel_bindings`, it’s likely we should carefully consider the various new interactions.

Since I’m not the most familiar with NTLM, I’ll cc other people: @TheBestTvarynka @RRRadicalEdward

inejge:
I believe that the quoted line is meant to be executed on the server when accepting the security context sent from the client. So, the channel bindings must have been already supplied by the client -- for which the current API lacks the method, hence my inquiry. But of course, let the other contributors chime in.

awakecoding:
For reference on channel bindings in the SSPI API just search "channel_bindings" in the FreeRDP source code: https://github.com/search?q=repo%3AFreeRDP%2FFreeRDP%20channel_bindings&type=code

I had forgotten that detail, but it looks like the channel binding is passed through an SSPI input buffer of type SECBUFFER_CHANNEL_BINDINGS.

However, I suppose you want a Rust API as well?

PR #479, opened 2025-08-15
feat(ntlm): add method to set the channel bindings for a session

inejge:
A while ago I opened #240, but didn't have time to thoroughly analyze the code and the spec in order to address the questions raised in the discussion. Now I've done the analysis. The principal question was:

> About the suggested change, however, my guess is that it was not intended for the `channel_bindings` field to be overridden by the user, as-is.

On the contrary, the user is the only party that can meaningfully set the channel bindings, since the point is to let the server know the user's idea of the server's identity, transmitted in a tamper-proof way, so that any discrepancy, indicative of an attempt of a relaying attack, can be detected and the session terminated.

Setting and sending the CBs is described in Sec. 3.1.5.2.1 of the NTLM spec, admittedly in a slightly confusing way, which may seem to suggest that the bindings are somehow part of the CHALLENGE message, but they are not: knowledge of CBs is strictly local to the client, therefore the client should send them in AUTHENTICATE.

Accordingly, the second commit in this PR removes the check for CBs in the CHALLENGE token, since it can't be there. The other check for CBs, on the server side (L287 in 78345a1fd875eac1b128cbcca1cefce81b339b26), is valid and should stay.

### DIFF
--- a/src/ntlm/mod.rs
+++ b/src/ntlm/mod.rs
@@ -351,12 +351,6 @@ impl Ntlm {
                 let input_token = SecurityBuffer::find_buffer(input, BufferType::Token)?;
                 let output_token = SecurityBuffer::find_buffer_mut(builder.output, BufferType::Token)?;
 
-                if let Ok(sec_buffer) =
-                    SecurityBuffer::find_buffer(builder.input.as_ref().unwrap(), BufferType::ChannelBindings)
-                {
-                    self.channel_bindings = Some(ChannelBindings::from_bytes(&sec_buffer.buffer)?);
-                }
-
                 client::read_challenge(self, input_token.buffer.as_slice())?;
 
                 client::write_authenticate(


CBenoit (commented):
Thank you!

issue: You need to run `cargo fmt` and verify the lints with `cargo clippy` before we can merge

inejge:
Uh yes, sorry about that. How do you prefer to have the changes published, a) rebase + fixup + force push, or b) simple push of the formatting commit?

CBenoit:
Either works, I'll squash the PR =)

inejge:
:+1:  Rebased and force pushed.

TheBestTvarynka (approved):
LGTM!

> Setting and sending the CBs is described in Sec. 3.1.5.2.1 of the NTLM spec, admittedly in a slightly confusing way, which may seem to suggest that the bindings are somehow part of the CHALLENGE message, but they are not: knowledge of CBs is strictly local to the client, therefore the client should send them in AUTHENTICATE.

Thank you for providing details and the spec section number. It helped me understand the desired behavior.

CBenoit (approved):
Thank you!

PR #491, opened 2025-09-02
feat(sspi): implement data sigining using scard

TheBestTvarynka:
Hi,
I implemented data signing using smart card. This is the last PR in smart card logon support.


mall refactoring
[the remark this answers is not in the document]
TheBestTvarynka: 🫨 
this file was created accidentally. I do not know even how.

src/utils.rs
```diff
@@ -14,6 +14,10 @@ pub fn string_to_utf16(value: impl AsRef<str>) -> Vec<u8> {
         .collect::<Vec<u8>>()
 }
 
+pub fn str_to_w_buff(data: &str) -> Vec<u16> {
+    data.encode_utf16().chain(std::iter::once(0)).collect()
+}
```
CBenoit, line 19: praise: It’s a good thing to use this helper at more places.

src/credssp/mod.rs
```diff
@@ -676,6 +676,7 @@ impl SspiImpl for SspiContext {
             SspiContext::Ntlm(ntlm) => {
                 let auth_identity = match builder.auth_data {
                     Some(Credentials::AuthIdentity(identity)) => Some(identity),
+                    #[allow(unreachable_patterns)]
```
CBenoit, line 679: question: If it’s unreachable, what about removing the whole branch?

suggestion: If you really want to keep this suppression, use `expect` so we get a warning when it does not apply anymore.

ffi/src/winscard/buf_alloc.rs
```diff
@@ -94,10 +95,17 @@ pub unsafe fn save_out_buf(out_buf: OutBuffer, p_buf: LpByte, pcb_buf: LpDword)
...
+                // The user is responsible for deallocating the buffer using the `SCardFreeMemory` function.
+                //
+                // The `p_buf` pointer is passed down to us from the outside world via FFI. Sometimes, this pointer
+                // can be misaligned. For example, if we deref the 0x170e56de4 pointer, then the Rust will panic with
+                // the 'misaligned pointer dereference: address must be a multiple of 0x8 but is...' error.
+                // We got this panic for the first time when connecting via FreeRDP + libsspi.dylib on macOS (arm64).
+                // So, we set the value using the `write_unaligned` function because it allows the pointer to be unaligned.
+                ptr::write_unaligned(p_buf, data.as_mut_ptr());
                 *pcb_buf = data.len().try_into()?;
             }
```
CBenoit, line 110: suggestion: Make sure the unsafe block is as small as possible, and there is no more than one unsafe operation per block

ffi/src/sspi/sec_winnt_auth_identity.rs
```diff
@@ -812,7 +800,7 @@ fn handle_smart_card_creds(mut username: Vec<u8>, password: Secret<Vec<u8>>) ->
         container_name: Some(string_to_utf16(key_container_name)),
         csp_name: string_to_utf16(csp_name),
         private_key_pem: None,
-        scard_type: SmartCardType::SystemProvided,
+        scard_type: SmartCardType::WindowsNative,
```
CBenoit, line 803: question: Is it fine even on Linux or macOS? Do we need conditional code to enable the emulated smart card based on the platform?

CBenoit (approved):
LGTM! Feel free to merge yourself

PR #495, opened 2025-09-05
refactor(sspi): fix lifetime anti-pattern in initialize_security_cont…

irvingoujAtDevolution:
…ext_impl

Replace &'a mut Struct<'a> anti-pattern with separate lifetimes for better flexibility. The previous signature forced all lifetimes to be the same, preventing generators from outliving their initialization parameters.

Changes:
- Separate context lifetime ('ctx), builder lifetime ('b), and generator lifetime ('g)
- Add lifetime bounds 'ctx: 'g and 'b: 'g to ensure safety
- Apply changes across all SSPI implementations (NTLM, Kerberos, Negotiate, PKU2U, CredSSP)
- Update trait definition in SspiImpl and FFI wrappers

This allows async generators to be used more flexibly without keeping initialization parameters alive for their entire lifetime.

### DIFF
--- a/ffi/src/sspi/sec_handle.rs
+++ b/ffi/src/sspi/sec_handle.rs
@@ -116,10 +116,14 @@ impl SspiImpl for SspiHandle {
         context.acquire_credentials_handle_impl(builder)
     }
 
-    fn initialize_security_context_impl<'a>(
-        &'a mut self,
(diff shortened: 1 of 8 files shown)


TheBestTvarynka (approved):
LGTM! Nice improvement

CBenoit (approved):
LGTM!

PR #499, opened 2025-09-09
build: optimize binary size for FFI builds

CBenoit:
Aim for good balance between size and performance.

- Baseline: 19M
- After: 8.9M

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -163,3 +163,49 @@ workspace = true
 [[example]]
 name = "kerberos"
 required-features = ["network_client"]
+
+[profile.ffi-production]
+inherits = "release"
+strip = "symbols"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -163,3 +163,49 @@ workspace = true
...
 name = "kerberos"
 required-features = ["network_client"]
+
+[profile.ffi-production]
+inherits = "release"
+strip = "symbols"
+lto = true
+opt-level = "z"
+codegen-units = 1
+panic = "abort"
```
CBenoit, line 173: Also set `panic = "abort"` because the code is typically written in a panic-resistant way I think. Does that sound good to you? @TheBestTvarynka

PR #506, opened 2025-09-29
feat(ffi): implement automatic WinSCard cache initialization

TheBestTvarynka:
Hi,
I improved WinSCard cache initialization in this PR.

cc @thenextman


CBenoit:
> It is a lot of work to revert and reopen PR, so I am going squash all commits before the last one and then we can marge&rebase this PR. How this sounds to you?

Sounds good to me!

PR #509, opened 2025-09-30
fix(winscard): general authenticate command validation

TheBestTvarynka:
Hi,
Recently I noticed that we have a failing test in the `winscard` crate. This PR fixes it.

_Note: I was able to reproduce the issue only when running `cargo test` with the `--all` arg._


CBenoit (approved):
LGTM!

CBenoit:
> _Note: I was able to reproduce the issue only when running `cargo test` with the `--all` arg._

Sounds like we do not run the sub-crates tests in the CI

PR #524, opened 2025-10-20
chore: get rid of `lazy_static`

RRRadicalEdward:
We still have it in the dependency tree because it's used in _num-bigint-dig_, which is still used by _picky-asn1-x509_.


CBenoit (approved):
🧹

PR #532, opened 2025-11-04
feat: implement `Default` for `SmartCardType`

allan2:
For use in [ironrdp_connector::credssp::CredSspSequence](https://github.com/Devolutions/IronRDP/pull/1028/files#diff-53ac0bab5ebaba1641334a689af83fc8bed484c7500d4aef5afa3e6cc11bd07cR130-R132)

### DIFF
--- a/src/auth_identity.rs
+++ b/src/auth_identity.rs
@@ -284,6 +284,14 @@ mod scard_credentials {
         WindowsNative,
     }
 
+    impl Default for SmartCardType {
+        fn default() -> Self {
+            Self::Emulated {
+                scard_pin: Secret::default(),
+            }
+        }
+    }
+
     /// Represents raw data needed for smart card authentication
     #[derive(Clone, Eq, PartialEq, Debug)]
     pub struct SmartCardIdentityBuffers {


CBenoit (approved):

src/auth_identity.rs
```diff
@@ -284,6 +284,14 @@ mod scard_credentials {
         WindowsNative,
     }
 
+    impl Default for SmartCardType {
+        fn default() -> Self {
+            Self::Emulated {
+                scard_pin: Secret::default(),
```
RRRadicalEdward: Is fine to use a default `scard_pin` here? I thought we need a user to pass it expictly to get it working. Will it work with a default one?
@TheBestTvarynka  Whay do you think?

TheBestTvarynka: It will definitely break the scard-based logon in IronRDP. The emulated smart card PIN should be retrieved from an environment variable or an external source. There is no other way for sspi-rs to know the smart card PIN code. Setting an empty PIN code by default is a bad idea.

I read the draft PR in IronRDP (https://github.com/Devolutions/IronRDP/pull/1028). The smart card should be determined by the environment variables. We can use the code from `ffi` as an example:
https://github.com/Devolutions/sspi-rs/blob/153900e58eb135b58c46ea22236f827cfffe2a0d/ffi/src/sspi/sec_winnt_auth_identity.rs#L487-L589

It can be implemented in a similar way in IronRDP.

cc @CBenoit

TheBestTvarynka: > I checked the RDP code, and we have the pin there:

In the context of emulated smart cards, we have two PINs: one is provided by the user during logon, and the second one is the actual smart card PIN. If we treat the PIN from the config as the smart card PIN, then any PIN provided by the user will be valid. I am not sure we want that 😅

The second PIN is needed to verify that the user typed a valid PIN code (expected PIN code).

https://github.com/Devolutions/sspi-rs/blob/153900e58eb135b58c46ea22236f827cfffe2a0d/src/auth_identity.rs#L270-L271

CBenoit: https://github.com/Devolutions/sspi-rs/pull/534

allan2: I agree!

PR #534, opened 2025-11-05
revert: implement `Default` for `SmartCardType`

CBenoit:
It actually does not make sense to implement Default for SmartCardType. A user-provided PIN must be set.

Release 0.18.2 was yanked.

Refs: 3555b377e8aeb5a8c376b2ecec9dcb1a2fd79560


CBenoit:
cc @allan2, please review this: https://github.com/Devolutions/sspi-rs/pull/532#discussion_r2493895938

PR #538, opened 2025-11-07
fix: pin leftover pre-release crypto crates

### LINKED ISSUE #537: Dependency stability: Release candidate dependencies causing frequent downstream breakages
Hi,

First, thank you for your work on this library. I wanted to bring attention to a recurring issue we're experiencing with the current dependency specifications.

### Current Situation
The library currently pins several dependencies to specific release *candidate* versions:
```toml
block-buffer = "=0.11.0-rc.5"
crypto-common = "=0.2.0-rc.4"
pkcs1 = "=0.8.0-rc.4"
pkcs8 = "=0.11.0-rc.7"
der = "=0.8.0-rc.9"
# ...
```

### Problem

These release candidate dependencies are causing frequent downstream breakages as they receive updates (roughly monthly). For example, `picky` crate is *currently* failing to compile with errors like:
```bash
 error[E0277]: the trait bound `R: crypto_bigint::rand_core::RngCore` is not satisfied
   --> crypto-primes-0.7.0-pre.4/src/hazmat/miller_rabin.rs:132:36
    |
132 |         let random = T::random_mod(rng, &range_nonzero)
    |                      ------------- ^^^ the trait `DerefMut` is not implemented for `R`
```

### Impact

While I understand this library is at version `0.18.1` and technically not promising API stability, Rust's cargo ecosystem typically maintains compatibility even for `0.x` versions. These RC dependency breaks are causing unexpected build failures for downstream users.

---

Would it make sense to make this library itself a release candidate (with `-rc` suffix) if it needs to depend on RC versions of other libraries? This would better signal to downstream users the potential for instability, and if there's a specific reason for needing these RC dependencies, it could be documented in the `Cargo.toml`.

Thank you

RRRadicalEdward:
The patch version upgrade is not allowed to bring breaking changes, but this rule doesn't work for an `rc` version. So we should pin the `rc` versions, to not allow _cargo_ update to a new `rc` automatically. 

Fixes #537.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1934,21 +1934,25 @@ checksum = "9b4f627cb1b25917193a259e49bdad08f671f8d9708acfd5fe0a8c1455d87220"
 
 [[package]]
 name = "picky"
-version = "7.0.0-rc.19"
+version = "7.0.0-rc.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "143fec6b7fa8ec91cbe1136f70ffa5de7e9627064da7bc3f7c0becc9b9493f8c"
+checksum = "4cdc52be663aebd70d7006ae305c87eb32a2b836d6c2f26f7e384f845d80b621"
 dependencies = [
  "base64",
  "block-buffer 0.11.0-rc.5",
  "block-padding",
  "crypto-bigint",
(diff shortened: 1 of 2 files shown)


CBenoit:
I’m currently merging the release PR, new version should be published soon!

PR #541, opened 2025-11-12
feat(kdc): add initial KDC implementation

TheBestTvarynka:
Hi,

I added a simple KDC implementation.

This KDC implementation will be used by the Devolutions-Gateway to do the Kerberos credentials-injection.

First, I planned to place it in the Devolutions-Gateway's repo, but then I noticed that it requires ASN1-related `picky-*` deps and I didn't want to add them to Devolution-Gateway.
I decided to placed it in the `crates/kdc`. This way, we will keep all Kerberos-related implementations in one place, with the identical dependency versions.

I added many comments and references to the RFC, so it should not be difficult to review this PR.

### DIFF
--- a/crates/kdc/src/error.rs
+++ b/crates/kdc/src/error.rs
@@ -109,7 +109,7 @@ impl KdcError {
                 name_string: ExplicitContextTag1::from(Asn1SequenceOf::from(vec![
                     KerberosStringAsn1::from(
                         IA5String::from_string(TGT_SERVICE_NAME.to_owned())
-                            .expect("TBT_SERVICE_NAME is valid KerberosString"),
+                            .expect("TGT_SERVICE_NAME is valid KerberosString"),
                     ),
                     realm,
                 ])),
--- a/crates/kdc/src/lib.rs
+++ b/crates/kdc/src/lib.rs
@@ -100,7 +100,7 @@ fn validate_request_from_and_till(
     // > If the requested expiration time minus the starttime (as determined above) is less than a site-determined minimum lifetime,
     // > an error message with code KDC_ERR_NEVER_VALID is returned.
     //
-    // We do not have a ticket minimum lifetime value configured, so we only check that the `end_time`` is after the `auth_time`.
+    // We do not have a ticket minimum lifetime value configured, so we only check that the `end_time` is after the `auth_time`.
     if end_time < auth_time {
         return Err(KdcError::NeverValid("end_time is earlier than auth_time".to_owned()));
     }
--- a/crates/kdc/src/tgs_exchange.rs
+++ b/crates/kdc/src/tgs_exchange.rs
@@ -9,9 +9,7 @@ use picky_krb::constants::etypes::{AES128_CTS_HMAC_SHA1_96, AES256_CTS_HMAC_SHA1
 use picky_krb::constants::key_usages::{
     TGS_REP_ENC_SESSION_KEY, TGS_REP_ENC_SUB_KEY, TGS_REQ_PA_DATA_AP_REQ_AUTHENTICATOR, TICKET_REP,
 };
-use picky_krb::constants::types::{
-    AS_REQ_MSG_TYPE, ENC_TGS_REP_PART_TYPE, PA_TGS_REQ_TYPE, TGS_REP_MSG_TYPE, TGS_REQ_MSG_TYPE,
-};
+use picky_krb::constants::types::{ENC_TGS_REP_PART_TYPE, PA_TGS_REQ_TYPE, TGS_REP_MSG_TYPE, TGS_REQ_MSG_TYPE};
 use picky_krb::crypto::CipherSuite;
 use picky_krb::data_types::{
     Authenticator, EncTicketPart, EncTicketPartInner, EncryptedData, PaData, PrincipalName, TicketInner,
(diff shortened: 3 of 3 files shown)


crates/kdc/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "kdc"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+picky-asn1 = { workspace = true, features = ["time_conversion"] }
+picky-asn1-der.workspace = true
+picky-asn1-x509.workspace = true
```
CBenoit: I’m really sorry to say that now, but we’ll stop using the picky crates soon in the future in favor of using the `der` crate from RustCrypto. Still, I think it’s fine to merge this code now that it’s here, but we should stop writing any new code using this infrastructure from now on. If it’s not too much work for you, you may consider changing that before we merge.

crates/kdc/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+picky-krb.workspace = true
```
CBenoit: For picky-krb it’s fine to keep it, but ultimately we’ll move it into sspi-rs, and migrate away from the picky infrastructure for serialization/deserialization.

crates/kdc/Cargo.toml
TheBestTvarynka: > If it’s not too much work for you, you may consider changing that before we merge.

To remove `picky-asn1-*` usage in the `kdc` implementation, we need to migrate the `picky-krb` crate to `der`-based encoding/decoding. It will significantly delay the credentials-injection feature.

I would like to address it in a separate PR. We need to refactor the `picky-krb` first.

crates/kdc/src/lib.rs
```diff
@@ -0,0 +1,186 @@
...
+    max_time_skew: u64,
+) -> Result<(OffsetDateTime, OffsetDateTime), KdcError> {
+    let now = OffsetDateTime::now_utc();
+    let max_time_skew = Duration::from_secs(max_time_skew);
+
+    let auth_time = if let Some(from) = from {
+        let from = OffsetDateTime::try_from(from.0.clone())
+            .map_err(|err| KdcError::NeverValid(format!("KdcReq::from time is not valid: {err}")))?;
+        // RFC (https://www.rfc-editor.org/rfc/rfc4120#section-3.1.3):
+        // > If the requested starttime is absent, indicates a time in the past,
```
CBenoit:
```suggestion
        // > If the requested start time is absent, indicates a time in the past,
```

CBenoit:
Excellent job, let’s just fix the minor typos etc before merging :+1:

crates/kdc/src/lib.rs
TheBestTvarynka: I copied this sentence directly from the RFC. Should I change it, though?

CBenoit: I think it will end up annoying if we use typos-cli at some point in the future 🤔

crates/kdc/src/error.rs
```diff
@@ -0,0 +1,223 @@
...
+            error_code: ExplicitContextTag6::from(KRB_ERR_GENERIC),
+            crealm: Optional::from(None),
+            cname: Optional::from(None),
+            realm: ExplicitContextTag9::from(realm.clone()),
+            sname: ExplicitContextTag10::from(PrincipalName {
+                name_type: ExplicitContextTag0::from(IntegerAsn1::from(vec![NT_SRV_INST])),
+                name_string: ExplicitContextTag1::from(Asn1SequenceOf::from(vec![
+                    KerberosStringAsn1::from(
+                        IA5String::from_string(TGT_SERVICE_NAME.to_owned())
+                            .expect("TBT_SERVICE_NAME is valid KerberosString"),
```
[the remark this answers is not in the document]
TheBestTvarynka: done

crates/kdc/src/ticket.rs
```diff
@@ -0,0 +1,165 @@
...
+        start_time: Optional::from(Some(ExplicitContextTag6::from(KerberosTime::from(
+            GeneralizedTime::from(auth_time),
+        )))),
+        end_time: ExplicitContextTag7::from(KerberosTime::from(GeneralizedTime::from(end_time))),
+        renew_till: Optional::from(None),
+        srealm: ExplicitContextTag9::from(realm),
+        sname: ExplicitContextTag10::from(sname),
+        // RFC (https://datatracker.ietf.org/doc/html/rfc4120#section-3.1.3):
+        // > ...It then formats a KRB_AS_REP message, copying the addresses in the request into the caddr of the response...
+        caadr: Optional::from(addresses.map(|addresses| ExplicitContextTag11::from(Asn1SequenceOf::from(addresses.0)))),
```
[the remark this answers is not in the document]
TheBestTvarynka: It needs to be fixed on the `picky-krb` side first. I will submit a PR soon

crates/kdc/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "kdc"
+version = "0.1.0"
+edition = "2024"
+
```
CBenoit: note: If you don’t want to publish the initial version right now, you should add publish = false here (and remove that later when publishing is intended)

PR #647, opened 2026-04-16
Add KDC connect timeout to prevent 75s hang on unreachable KDC

thenextman:
`TcpStream::connect()` and `UdpSocket::recv()` had no timeout when contacting the KDC, causing authentication to hang until the OS gave up. On macOS this is controlled by net.inet.tcp.keepinit (default 75s); the duration will differ on other platforms but the unbounded  block exists on all of them.

This is most visible after sleep/wake or dock/undock, where the KDC is temporarily unreachable during network re-initialization. The full OS timeout elapses before Negotiate falls back to NTLM.                                                                          
  
Fix: use `TcpStream::connect_timeout()` (5s, matching Windows Kerberos behaviour) with a `to_socket_addrs()` loop to try all resolved addresses, and `set_read_timeout()` on the UDP socket. Failed attempts are logged at DEBUG with the address and error.

### DIFF
--- a/src/network_client.rs
+++ b/src/network_client.rs
@@ -54,7 +54,8 @@ pub trait NetworkClient: Send + Sync {
 #[cfg(feature = "network_client")]
 pub mod reqwest_network_client {
     use std::io::{Read, Write};
-    use std::net::{IpAddr, Ipv4Addr, TcpStream, UdpSocket};
+    use std::net::{IpAddr, Ipv4Addr, TcpStream, ToSocketAddrs, UdpSocket};
+    use std::time::Duration;
 
     use byteorder::{BigEndian, ReadBytesExt};
     use url::Url;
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
LGTM!

thenextman:
@CBenoit If and when you have time, please take a look at the changes for error handling in the UDP case. I think it's correct.

src/network_client.rs
```diff
@@ -94,14 +119,28 @@ pub mod reqwest_network_client {
...
-            let n = udp_socket.recv(&mut buf)?;
+            let n = udp_socket.recv(&mut buf).map_err(|e| {
+                let msg = format!("{e:?}");
+                match e.kind() {
+                    std::io::ErrorKind::TimedOut | std::io::ErrorKind::WouldBlock => {
+                        debug!(error = %msg, "KDC UDP recv timed out");
+                        Error::new(ErrorKind::NoAuthenticatingAuthority, msg)
+                    }
+                    _ => Error::new(ErrorKind::NoAuthenticatingAuthority, msg),
+                }
```
CBenoit: I think it’s a bit too complex for what it really is:

```suggestion
                Error::new(ErrorKind::NoAuthenticatingAuthority, msg)
```

We don’t need an extra debug! because the error will be logged somewhere anyway (the `msg` we insert into `Error::new`).

CBenoit: Basically the same as above:

```rust
            udp_socket
                .send_to(data, addr) // <-- But it’s `recv`
                .map_err(|e| Error::new(ErrorKind::NoAuthenticatingAuthority, format!("{e:?}")))?;
```

CBenoit (approved):
Looks good to me, but I think the map_err introduced to fix the problem is a bit too complex. I don’t think it does any harm though

PR #660, opened 2026-05-05
fix!: remove redundant PKCS#1 padding when signing hashes

rhammonds-teleport:
Over at Teleport, we've noticed some errors from the winscard crate during logon with NLA:
```
CredSSP InternalError: Error while using a smart card: Error{ InternalError: Error: an unexpected RsaError happened: input must be hashed }
```

`Pkcs1v15Sign::new` already handles PKCS#1 padding internally and checks that the input vector matches the expected length of a SHA1 hash. Since the smart card implementation is modifying the input vector with `encode_digest`, this check fails resulting in the error above.

### DIFF
--- a/src/smartcard.rs
+++ b/src/smartcard.rs
@@ -167,7 +167,7 @@ impl SmartCard {
         match &mut self.smart_card_type {
             SmartCardApi::PivEmulated(scard) => {
                 scard.verify_pin(self.pin.as_ref())?;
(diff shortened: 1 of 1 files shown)


TheBestTvarynka (commented):
🤔 
I am wondering how we missed it. Maybe they changed the `Pkcs1v15Sign` behavior in the latest versions

CBenoit (approved):
Thank you! I’ll publish a hotfix patch!

PR #676, opened 2026-05-25
feat(ffi): add support for SECPKG_ATTR_NAMES

### LINKED ISSUE #169: Support QueryContextAttributes SECPKG_ATTR_NAMES
Would be great to support `SECPKG_ATTR_NAMES` with [QueryContextAttributesW](https://learn.microsoft.com/en-us/windows/win32/api/sspi/nf-sspi-querycontextattributesw) to retrieve the client principal name from the authentication context. This is mostly just a helpful function to easily debug/log the username of the other end.

TheBestTvarynka:
Hi,
I got some time between code reviews and decided to address this small old issue.

closes #169

### DIFF
--- a/ffi/src/sspi/sec_handle.rs
+++ b/ffi/src/sspi/sec_handle.rs
@@ -17,7 +17,7 @@ use sspi::ntlm::NtlmConfig;
 use sspi::{
     CertContext, ClientRequestFlags, ConnectionInfo, Credentials, CredentialsBuffers, DataRepresentation, Error,
     ErrorKind, Kerberos, Negotiate, NegotiateConfig, Ntlm, PackageInfo, Result, Secret, Sspi, SspiImpl, StreamSizes,
-    U16CString, kerberos, negotiate, ntlm, pku2u,
+    U16CString, Utf16String, Utf16StringExt, kerberos, negotiate, ntlm, pku2u,
 };
 #[cfg(target_os = "windows")]
 use windows::Win32::Security::Cryptography::{
@@ -1033,6 +1033,10 @@ unsafe fn query_context_attributes_common(
                     libc::malloc(session_key_len)
                 }.cast::<u8>();
 
+                if buf.is_null() {
+                    return ErrorKind::InsufficientMemory.to_u32().unwrap();
+                }
+
                 // SAFETY:
                 // - `session_key.as_ref().as_ptr()` pointer is valid for reads of `session_key_len` bytes,
                 //   because it is Rust-allocated vector of u8 values with size of `session_key_len`.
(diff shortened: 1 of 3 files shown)


ffi/src/sspi/sec_handle.rs
```diff
@@ -1049,7 +1052,74 @@ unsafe fn query_context_attributes_common(
...
+            SECPKG_ATTR_NAMES => {
+                let sspi_names = try_execute!(sspi_context.query_context_names());
+
+                if is_wide {
+                    let sec_pkg_names = p_buffer.cast::<SecPkgContextNamesW>();
+                    let wide_username: Vec<u16> = sspi_names.username.account_name().encode_utf16().chain(std::iter::once(0)).collect();
+                    let username_len = wide_username.len();
+
+                    // SAFETY: Memory allocation is safe.
+                    let buf = unsafe {
```
RRRadicalEdward, line 1069: Shouldn't you check that `buf` is not null?

ffi/src/sspi/sec_handle.rs
```diff
@@ -1049,7 +1052,74 @@ unsafe fn query_context_attributes_common(
...
+                    }
+
+                    return 0;
+                } else {
+                    let sec_pkg_names = p_buffer.cast::<SecPkgContextNamesA>();
+
+                    let username = sspi_names.username.account_name();
+
+                    // SAFETY: Memory allocation is safe.
+                    let buf = unsafe {
```
RRRadicalEdward, line 1101: The same here

ffi/src/sspi/sec_handle.rs
```diff
@@ -1049,7 +1052,74 @@ unsafe fn query_context_attributes_common(
 
                 return 0;
             }
-            _ => {}
+            SECPKG_ATTR_NAMES => {
+                let sspi_names = try_execute!(sspi_context.query_context_names());
+
+                if is_wide {
+                    let sec_pkg_names = p_buffer.cast::<SecPkgContextNamesW>();
+                    let wide_username: Vec<u16> = sspi_names.username.account_name().encode_utf16().chain(std::iter::once(0)).collect();
```
RRRadicalEdward: Use `Utf16String::from_str(...).into_vec_with_nul()` ?

ffi/src/sspi/sec_handle.rs
TheBestTvarynka: Oh, yes. Fixed

ffi/src/sspi/sec_handle.rs
TheBestTvarynka: Fixed

ffi/src/sspi/sec_handle.rs
TheBestTvarynka: Correct. Forgot about this API

ffi/src/sspi/sec_handle.rs
```diff
@@ -1049,7 +1052,74 @@ unsafe fn query_context_attributes_common(
...
+                    // SAFETY: `sec_pkg_names` is non-null because it was cast from a non-null `p_buffer`.
+                    unsafe {
+                        (*sec_pkg_names).user_name = buf;
+                    }
+
+                    return 0;
+                } else {
+                    let sec_pkg_names = p_buffer.cast::<SecPkgContextNamesA>();
+
+                    let username = sspi_names.username.account_name();
```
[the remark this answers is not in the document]
TheBestTvarynka: Fixed by replacing `.account_name()` with `.inner()`. Also added a comment in the code explaining this decision.

ffi/src/sspi/sec_handle.rs
```diff
@@ -2334,4 +2404,59 @@ mod tests {
...
+
+        let status =
+            unsafe { QueryContextAttributesA(&mut sec_handle, SECPKG_ATTR_NAMES, ptr::from_mut(&mut names_a).cast()) };
+        assert_eq!(status, 0);
+        assert!(!names_a.user_name.is_null());
+
+        let username_a = unsafe { CStr::from_ptr(names_a.user_name) }
+            .to_str()
+            .expect("username should be valid UTF-8");
+        println!("Username (ANSI): {}", username_a);
```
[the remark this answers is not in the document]
TheBestTvarynka: > but never asserts the actual decoded username matches the value seeded into

The fake credentials are hardcoded in the `fake_client` function. We cannot obtain them from the `Kerberos` object. The only way is to export fake credentials from `sspi`, but I do not want to increase the exported API. If I hardcode credentials in the test, then they may change in the future, and we will get a false positive. Also, the main purpose of this test is to check for UB and memory-related mistakes.

TheBestTvarynka:
@CBenoit, I addressed all comments

CBenoit (approved):
Thanks!

PR #679, opened 2026-05-26
feat(ffi): generate dotnet bindings with csbindgen

philippelatulippe:
Currently the Devolutions.Sspi nuget only contains a native DLL and no bindings, so users of the nuget need to manually define method signatures in their project. This is prone to silent breakage (and likely memory corruption) if the signatures in the DLL change.

This PR adds a build dependency on [csbindgen](https://github.com/Cysharp/csbindgen) to add generated bindings to the C# project in ffi/dotnet.

Open questions:

* I've set it up to generate one big C# file. I could break it up into multiple classes: dpapi, sspi, winscard.
* The cargo lockfile has unexpected changes, is it an issue on my end or did someone forget to commit their lockfile?
* I committed the generated C# source so that you can see it, but perhaps the repo has a different policy?


ffi/build.rs
```diff
@@ -128,4 +128,42 @@ fn main() {
...
+                .input_extern_file("src/winscard/scard_context.rs")
+                .input_extern_file("src/winscard/scard.rs")
+                .input_extern_file("../crates/ffi-types/src/winscard/functions.rs")
+                .input_extern_file("../crates/ffi-types/src/winscard/mod.rs")
+                .input_extern_file("../crates/ffi-types/src/common.rs");
+        }
+
+        csbindgen_builder
+            .csharp_namespace("Devolutions.Sspi")
+            .csharp_class_name("SspiRs")
```
CBenoit: My preference would be for `Sspi` instead of `SspiRs`: `-rs` is the suffix of the repository, but we don’t really name anything in the code with `_rs` or `Rs` otherwise

ffi/Cargo.toml
```diff
@@ -62,6 +62,9 @@ bitflags = { workspace = true, optional = true }
 [target.'cfg(not(any(target_os="macos", target_os="ios", windows)))'.dependencies]
 sspi = { path = "..", default-features = false, features = ["dns_resolver"] }
 
+[build-dependencies]
+csbindgen = "1.9.6"
```
CBenoit: Overconstrained

```suggestion
csbindgen = "1.9"
```

CBenoit (commented):
This is looking good to me. `csbindgen` seems to be the appropriate tool for a low-effort, low-level helper library. I think it’s fine to have a single file, I don’t think we’ll really go read in details anyway. Not sure for the Cargo.lock, but indeed we forgot to use `--locked` in the CI, so it’s possible to forget.

CBenoit:
Also cc @PavloMyroniuk-apriorit 
If you have any opinion here?

philippelatulippe:
Richard mentioned that my usage of [DllImport] isn't cross-platform, and that simply removing ".dll" isn't sufficient on iOS. I'll look into implementing a DllImportResolver in the template (which is supported by csbindgen) or and #if statement for the name (only work if we do a separate build for iOS, not sure if that's the case)

CBenoit:
If it helps, that’s what we have in picky-rs: https://github.com/Devolutions/picky-rs/blob/8a35c639565832ad745982120d39cdc385738641/ffi/dotnet/Devolutions.Picky/Generated/RawJwtSig.cs#L23-L27

Is it something you can customize in csbindgen?

thenextman:
@CBenoit The main difference with picky is that it requires an iOS specific build _anyway_ because we need to decorate reverse p/invoke with the proper attribute. sspi-rs doesn't need that; although the same approach would work and future proof us for future reverse callbacks :) 

A custom DllImportResolver which just rewrites the library name to an rpath if we detect we're running on iOS as @philippelatulippe suggested should also work and might be a bit simpler. Really depends on csbindgen.

TheBestTvarynka (commented):
LGTM. The build config is clean and simple. I like the approach

philippelatulippe:
Just pushed:
* C# class is now `Sspi`
* Relaxed csbindgen version in cargo.toml
* In the C# library, added a custom DLL loader for iOS
* Added an optional C# unit test project that pulls a locally-built nuget and checks that the methods of the native library can properly be called. This was to help me test the changes on manual platforms more easily.
* Testing a nuget package isn't natively supported, instead I created an extra build configuration for the library's csproj that sets a package version based on the current time. The test csproj has a hardcoded relative package source that points to the folder where nugets are built

ffi/dotnet/Devolutions.Sspi/Devolutions.Sspi.csproj
```diff
@@ -3,14 +3,19 @@
   <PropertyGroup>
     <Company>Devolutions</Company>
     <Description>Portable Rust SSPI library</Description>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFramework>net6.0</TargetFramework>
     <LangVersion>latest</LangVersion>
```
CBenoit, line 7: issue?

I don’t remember this was changed in the initial patch, but it seems to me that this should be its own PR so its not a "silent" change, especially as this may be a problem downstream 🤔

I’m pretty sure we don’t have anything lower than net6.0, but I’m pinging @thenextman just in case 👀

CBenoit (commented):
I’m happy with the general direction, and I really appreciate the small smoke test!

I want to defer the final approval to @thenextman as I’m not sure about the new changes in the .csproj files, but everything else is looking good to me!

ffi/dotnet/Devolutions.Sspi/Devolutions.Sspi.csproj
philippelatulippe: I totally forgot to mention it, I had to bump up the target version in order to use `DllImportResolver`, which was introduced in .NET Core 3.0 and isn't in netstandard. Richard suggested I go with net6.0. I realize this will need a mention in the release notes since it might be a breaking change if anyone is still on .NET Framework. How should I proceed?

thenextman: Two approaches I can suggest:

- Keep the bump and leave everything else as-is. But yes, maybe we should make that a separate commit.
- Multitarget netstandard2.0 and net-6.0 (or other LTS target). Gate the assembly resolver behind an `#IF` directive.

In the first case, we cut support broadly for anyone not on a .NET LTS target. But c'mon, .NET 6 was released five years ago.

In the second case, we keep broad support for the native library. The downside is that iOS users on a netstandard2.0-equivalent framework can't use the bindings (although they can still use the native library).

thenextman (approved):
I approve for velocity, it looks good to me; there is just the unresolved question regarding the TFM bump. I made a comment in-line for that.

ffi/dotnet/Devolutions.Sspi/Devolutions.Sspi.csproj
CBenoit: For the sake of simplicity and given the age of net6 I would lean towards dropping netstandard2

ffi/dotnet/Devolutions.Sspi/Devolutions.Sspi.csproj
```diff
@@ -3,14 +3,19 @@
...
     <Nullable>enable</Nullable>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <SuppressDependenciesWhenPacking>false</SuppressDependenciesWhenPacking>
+    <Configurations>Debug;Release;PkgTest</Configurations>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)'=='PkgTest'">
+    <Version>3000.0.0-$([System.DateTime]::Now.ToString('yyyyMMddHHmm'))</Version>
   </PropertyGroup>
```
CBenoit, line 21: nit: Also it may be good to document the workaround with inline comments. At a glance it's not obvious why we are doing something convoluted here.

philippelatulippe: Good idea, I added a comment

ffi/build.rs
philippelatulippe: Done

ffi/Cargo.toml
philippelatulippe: Done!

ffi/dotnet/Devolutions.Sspi/Devolutions.Sspi.csproj
philippelatulippe: I changed the target to net6.0 in PR #682 which is now merged

PR #680, opened 2026-05-29
fix(sspi): disable `mechListMIC` for NTLM guest logon in negotiate module

TheBestTvarynka:
Hi,
This PR fixes another issue that is related to #640. The regular SPNEGO NTLM scenario requires `mechListMIC` exchange except for the NTLM guest logon.

When the user tries to log on under the `/GUEST` account, the `sspi-rs` will not send the MIC token.

I also added tests to cover a new behavior.


CBenoit (approved):
Nothing to flag from my side! LGTM! Thank you 🚀

PR #682, opened 2026-06-01
build(ffi): Change C# bindings' target framework to net6.0

philippelatulippe:
Drops support for netstandard2.0 and therefore the old .NET Framework.

Paves the way for PR #679


CBenoit (approved):
Thanks!

PR #688, opened 2026-06-11
fix(credssp): complete the SPNEGO mechListMIC exchange in CredSspServer

### LINKED ISSUE #687: CredSspServer (Negotiate/NTLM) cannot complete a handshake with sspi's own CredSspClient: pubKeyAuth expected one leg too early
## Summary

Pairing `CredSspClient` with `CredSspServer`, both in `Negotiate` mode with NTLM pinned, never completes: the server fails on the client's AUTHENTICATE leg with

```
Error { error_type: InvalidToken, description: "CredSSP server expected an encrypted public key", nstatus: None }
```

The two sides disagree about **which leg carries `pubKeyAuth`**:

- The **client** (per the MS-SPNG rule that a MIC'd NTLM AUTHENTICATE makes the SPNEGO `mechListMIC` exchange mandatory — and `Negotiate` always includes the NTLM MIC, `src/negotiate/mod.rs`) returns `ContinueNeeded` from the AUTHENTICATE leg and **defers `pubKeyAuth`** until it has received and verified the server's `mechListMIC` (`src/negotiate/client.rs`; `CredSspClient` only attaches `pubKeyAuth` when the ISC status is `Ok`).
- The **server**'s Negotiate acceptor verifies the client's `mechListMIC` on the AUTHENTICATE leg and reports `Ok` immediately (`src/negotiate/server.rs`); `CredSspServer`'s `Ok`/`CompleteNeeded` arm then requires `pub_key_auth` **in that same TSRequest** and errors — additionally discarding the output token (its own `mechListMIC` response) via `ts_request.nego_tokens = None` (`src/credssp/mod.rs`).

So the client sends `pubKeyAuth` one leg later than the server demands it. The client's behavior appears to be the correct one: real Windows RDP servers complete this exact client flow (verified end-to-end against a live Windows host), answering the AUTHENTICATE leg with their own `mechListMIC` and accepting `pubKeyAuth` on the following leg.

Notably, the in-repo client/server tests (`tests/sspi/client_server/credssp.rs`) only pair `Ntlm↔Ntlm` (no SPNEGO, no MIC exchange — works) and `Negotiate↔Negotiate` for Kerberos, so this NTLM-inside-Negotiate pairing is never exercised.

## Environment

- sspi `0.21.0` (latest on crates.io), default features
- Windows 11, Rust stable
- No network, no KDC — pure in-memory `TsRequest` ping-pong

## Reproduction (self-contained)

```rust
use std::io;

use sspi::credssp::{
    ClientMode, ClientState, CredSspClient, CredSspMode, CredSspServer, CredentialsProxy,
    ServerMode, ServerState, TsRequest,
};
use sspi::generator::GeneratorState;
use sspi::ntlm::NtlmConfig;
use sspi::{AuthIdentity, Credentials, NegotiateConfig, Secret, Username};

const PUBLIC_KEY: &[u8] = &[0xAB; 64];

struct SingleUser(AuthIdentity);

impl CredentialsProxy for SingleUser {
    type AuthenticationData = AuthIdentity;
    fn auth_data_by_user(&mut self, username: &Username) -> io::Result<AuthIdentity> {
        Ok(AuthIdentity { username: username.clone(), password: self.0.password.clone() })
    }
    fn auth_data(&mut self) -> io::Result<Vec<AuthIdentity>> {
        Ok(vec![self.0.clone()])
    }
}

fn run(label: &str, client_mode: ClientMode, server_mode: ServerMode) {
    let identity = AuthIdentity {
        username: Username::parse("test_user").unwrap(),
        password: Secret::from("test_password".to_owned()),
    };
    let mut client = CredSspClient::new(
        PUBLIC_KEY.to_vec(),
        Credentials::AuthIdentity(identity.clone()),
        CredSspMode::WithCredentials,
        client_mode,
        "TERMSRV/localhost".to_owned(),
    )
    .unwrap();
    let mut server =
        CredSspServer::new(PUBLIC_KEY.to_vec(), SingleUser(identity), server_mode).unwrap();

    let mut ts_request = TsRequest::default();
    for leg in 1..=8 {
        let client_state = match client.process(std::mem::take(&mut ts_request)).start() {
            GeneratorState::Completed(result) => result.unwrap(),
            GeneratorState::Suspended(_) => unreachable!(),
        };
        let to_server = match client_state {
            ClientState::ReplyNeeded(ts) | ClientState::FinalMessage(ts) => ts,
        };
        println!(
            "[{label}] leg {leg}: client -> server  negoTokens={} pubKeyAuth={} authInfo={}",
            to_server.nego_tokens.is_some(),
            to_server.pub_key_auth.is_some(),
            to_server.auth_info.is_some(),
        );
        match server.process(to_server).start() {
            GeneratorState::Completed(Ok(ServerState::ReplyNeeded(reply))) => ts_request = reply,
            GeneratorState::Completed(Ok(ServerState::Finished(id))) => {
                println!("[{label}] FINISHED: server authenticated {:?}\n", id.username);
                return;
            }
            GeneratorState::Completed(Err(e)) => {
                println!("[{label}] SERVER ERROR on leg {leg}: {:?}\n", e.error);
                return;
            }
            GeneratorState::Suspended(_) => unreachable!(),
        }
    }
    println!("[{label}] did not converge within 8 legs\n");
}

fn main() {
    // Control: the pairing the in-repo tests cover — completes.
    run(
        "ntlm-ntlm",
        ClientMode::Ntlm(NtlmConfig::default()),
        ServerMode::Ntlm(NtlmConfig::default()),
    );
    // Broken: both sides Negotiate with NTLM pinned — the mode RDP clients need,
    // since Windows RDP servers reject bare NTLM inside CredSSP.
    run(
        "negotiate-negotiate",
        ClientMode::Negotiate(NegotiateConfig::new(
            Box::new(NtlmConfig::default()),
            Some("ntlm,!kerberos,!pku2u".to_owned()),
            "CLIENT".to_owned(),
        )),
        ServerMode::Negotiate(NegotiateConfig::new(
            Box::new(NtlmConfig::default()),
            Some("ntlm,!kerberos,!pku2u".to_owned()),
            "SERVER".to_owned(),
        )),
    );
}
```

### Output

```
[ntlm-ntlm] leg 1: client -> server  negoTokens=true pubKeyAuth=false authInfo=false
[ntlm-ntlm] leg 2: client -> server  negoTokens=true pubKeyAuth=true authInfo=false
[ntlm-ntlm] leg 3: client -> server  negoTokens=false pubKeyAuth=false authInfo=true
[ntlm-ntlm] FINISHED: server authenticated Username { value: "test_user", format: DownLevelLogonName, sep_idx: None }

[negotiate-negotiate] leg 1: client -> server  negoTokens=true pubKeyAuth=false authInfo=false
[negotiate-negotiate] leg 2: client -> server  negoTokens=true pubKeyAuth=false authInfo=false
[negotiate-negotiate] SERVER ERROR on leg 2: Error { error_type: InvalidToken, description: "CredSSP server expected an encrypted public key", nstatus: None }
```

## Expected behavior

When the acceptor still has a `mechListMIC` response to deliver, `CredSspServer` should answer the AUTHENTICATE leg with `ServerState::ReplyNeeded` carrying that token (instead of dropping it) and accept `pubKeyAuth` on the **following** leg — matching both its own client's behavior and what Windows servers do.

## Impact / context

Found while writing loopback integration tests for an RDP client that uses `CredSspClient` in `Negotiate` mode (required for real Windows targets): there is currently no way to stand up an sspi-based mock CredSSP server for it. Possibly related to the recent client-side mechListMIC work in #640 / #670, but this is specifically the **server** half of the exchange.

kihyun1998:
Closes #687.

## Problem

`CredSspServer` cannot authenticate sspi's own `CredSspClient` when both sides run NTLM inside SPNEGO (`ClientMode::Negotiate` / `ServerMode::Negotiate` — the pairing RDP requires, since Windows RDP servers reject bare NTLM inside CredSSP). The handshake fails on the AUTHENTICATE leg with:

```
Error { error_type: InvalidToken, description: "CredSSP server expected an encrypted public key", nstatus: None }
```

Because the client always includes an NTLM MIC, the SPNEGO `mechListMIC` exchange is mandatory ([MS-SPNG 3.3.5.5](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-spng/f377a379-c24f-4a0f-a3eb-0d835389e28a) / RFC 4178 §5), so the client defers `pubKeyAuth` until it has received and verified the server's final SPNEGO token. The server side, however, dropped that final token (`ts_request.nego_tokens = None`) at the moment its acceptor completed, and demanded `pubKeyAuth` in the same TSRequest — one leg earlier than its own client sends it.

## Fix

In `CredSspServer`'s `NegoToken` state, when the acceptor completes (`Ok`/`CompleteNeeded`) but the incoming TSRequest carries **no** `pubKeyAuth` and the acceptor produced a final output token:

- reply `ServerState::ReplyNeeded` carrying that final SPNEGO token (accept-completed + `mechListMIC`) instead of dropping it, and
- accept `pubKeyAuth` on the **following** leg (`awaiting_pub_key_auth` flag; calling the acceptor again on that leg would hit `NegotiateState::Ok → OutOfSequence`).

This matches the client's behavior and what Windows RDP servers do on the wire. Pairings where `pubKeyAuth` arrives together with the last nego token (bare NTLM, Kerberos) take the unchanged single-leg path — the new branch only activates when `pubKeyAuth` is absent *and* a final token exists. The `pubKeyAuth` decrypt/encrypt sequence is extracted into `exchange_pub_key_auth` to avoid duplicating it across the two legs.

## Resulting wire flow (Negotiate-NTLM)

```
before: NEGOTIATE → CHALLENGE → AUTHENTICATE+mechListMIC → ❌ InvalidToken
after:  NEGOTIATE → CHALLENGE → AUTHENTICATE+mechListMIC → accept-completed+mechListMIC
        → pubKeyAuth → pubKeyAuth (server) → authInfo → Finished ✅
```

## Tests

- New `credssp_negotiate_ntlm` client↔server test pins the four-client-leg handshake — it fails with the pre-fix error on master and passes with this change. (The suite previously paired only `Ntlm↔Ntlm` and `Negotiate↔Negotiate`-Kerberos, so this path was never exercised.)
- Existing `credssp_ntlm` (3 legs, unchanged) and `credssp_kerberos` (single-leg `pubKeyAuth`, unchanged) still pass — `cargo test --test sspi --features "network_client,__test-data"`: 19 passed; `cargo test -p sspi --lib`: 230 passed.

Found while building an RDP client that drives `CredSspClient` in Negotiate mode against real Windows hosts (which complete this exact client flow) and attempting to stand up an sspi-based loopback CredSSP server for integration tests.

## CI

`f8959ca` adds `__test-data` to the three root-manifest test-matrix rows, so the `client_server` suite (gated on `all(network_client, __test-data)`) actually runs in CI from now on — it previously never did, which is how #687 shipped unnoticed. The suite is hermetic (`KdcMock` / `NetworkClientMock`); verified locally with the exact win-row feature combo and the shared minimal combo, all green.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,16 +101,18 @@ jobs:
           - manifest: crates/dpapi-native-transport/Cargo.toml
             crate-name: dpapi-native-transport
 
-          # Per-OS feature overrides for specific manifests
+          # Per-OS feature overrides for specific manifests.
+          # `__test-data` enables the `tests/sspi/client_server` suite (gated on
+          # `all(network_client, __test-data)`), which otherwise never runs in CI.
           - os: win
             manifest: Cargo.toml
-            additional-args: --features network_client,dns_resolver,scard,tsssp
+            additional-args: --features network_client,dns_resolver,scard,tsssp,__test-data
           - os: osx
             manifest: Cargo.toml
-            additional-args: --features network_client,scard
+            additional-args: --features network_client,scard,__test-data
           - os: linux
             manifest: Cargo.toml
-            additional-args: --features network_client,dns_resolver,scard
+            additional-args: --features network_client,dns_resolver,scard,__test-data
 
           - os: win
             manifest: ffi/Cargo.toml
--- a/src/credssp/mod.rs
+++ b/src/credssp/mod.rs
@@ -513,7 +513,12 @@ impl<C: CredentialsProxy<AuthenticationData = AuthIdentity> + Send> CredSspServe
             client_nonce,
             peer_version,
         )?;
-        context.encrypt_public_key(self.public_key.as_ref(), EndpointType::Server, client_nonce, peer_version)
+        context.encrypt_public_key(
+            self.public_key.as_ref(),
+            EndpointType::Server,
+            client_nonce,
+            peer_version,
+        )
     }
 
     #[instrument(fields(state = ?self.state), skip_all)]
@@ -621,10 +626,8 @@ impl<C: CredentialsProxy<AuthenticationData = AuthIdentity> + Send> CredSspServe
                         ts_request
                     );
                     let client_nonce = ts_request.client_nonce;
-                    let pub_key_auth = try_cred_ssp_server!(
-                        self.exchange_pub_key_auth(pub_key_auth, &client_nonce),
-                        ts_request
-                    );
+                    let pub_key_auth =
+                        try_cred_ssp_server!(self.exchange_pub_key_auth(pub_key_auth, &client_nonce), ts_request);
                     ts_request.nego_tokens = None;
                     ts_request.pub_key_auth = Some(pub_key_auth);
 
--- a/src/credssp/mod.rs
+++ b/src/credssp/mod.rs
@@ -454,6 +454,10 @@ pub struct CredSspServer<C: CredentialsProxy<AuthenticationData = AuthIdentity>>
     credentials_handle: Option<CredentialsBuffers>,
     ts_request_version: u32,
     context_config: Option<ServerMode>,
+    /// Set when the security context completed but the final SPNEGO token (accept-completed +
+    /// `mechListMIC`) still had to be sent to the client: the client cannot send `pubKeyAuth`
+    /// until it has verified that token, so `pubKeyAuth` arrives on the *next* leg.
+    awaiting_pub_key_auth: bool,
 }
 
 impl<C: CredentialsProxy<AuthenticationData = AuthIdentity> + Send> CredSspServer<C> {
@@ -466,6 +470,7 @@ impl<C: CredentialsProxy<AuthenticationData = AuthIdentity> + Send> CredSspServe
             credentials_handle: None,
             ts_request_version: TS_REQUEST_VERSION,
             context_config: Some(client_mode),
+            awaiting_pub_key_auth: false,
         })
     }
 
@@ -483,7 +488,32 @@ impl<C: CredentialsProxy<AuthenticationData = AuthIdentity> + Send> CredSspServe
             credentials_handle: None,
             ts_request_version,
             context_config: Some(client_mode),
+            awaiting_pub_key_auth: false,
         })
+    }
+
+    /// Decrypts and verifies the client's `pubKeyAuth`, returning the server-side `pubKeyAuth`
+    /// to send back.
+    fn exchange_pub_key_auth(
+        &mut self,
+        pub_key_auth: Vec<u8>,
+        client_nonce: &Option<[u8; NONCE_SIZE]>,
+    ) -> crate::Result<Vec<u8>> {
+        let peer_version = self
+            .context
+            .as_ref()
+            .unwrap()
+            .peer_version
+            .expect("an decrypt public key server function cannot be fired without any incoming TSRequest");
+        let context = self.context.as_mut().unwrap();
+        context.decrypt_public_key(
+            self.public_key.as_ref(),
+            pub_key_auth.as_ref(),
+            EndpointType::Server,
+            client_nonce,
+            peer_version,
+        )?;
+        context.encrypt_public_key(self.public_key.as_ref(), EndpointType::Server, client_nonce, peer_version)
     }
 
     #[instrument(fields(state = ?self.state), skip_all)]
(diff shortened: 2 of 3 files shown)


kihyun1998:
One observation from reading the CI config while preparing this PR: the `client_server` test module is gated on `all(feature = "network_client", feature = "__test-data")`, but no row of the CI test matrix enables `__test-data` — so the client↔server pairing tests (including `credssp_ntlm`, `credssp_kerberos`, and the regression test added here) currently only run locally, never in CI. That is likely how #687 went unnoticed.

Happy to add `__test-data` to the matrix rows in this PR if you'd like — or leave it to a separate change if you prefer to keep this one scoped to the fix.

kihyun1998:
Went ahead and added it in f8959ca (kept as a separate commit so it's easy to drop if you'd rather scope this PR to the fix). Verified locally with the win-row combo `network_client,dns_resolver,scard,tsssp,__test-data` — the suite is hermetic and green.

tests/sspi/client_server/credssp.rs
```diff
@@ -109,6 +109,54 @@ fn credssp_ntlm() {
     run_credssp(&mut client, &mut server, &auth_identity, &mut network_client);
 }
 
+#[test]
+fn credssp_negotiate_ntlm() {
```
TheBestTvarynka, line 113: praise: good test 💟

.github/workflows/ci.yml
TheBestTvarynka, line 1: This change should be in a separate PR. We try not to mix CI and library code changes

src/credssp/mod.rs
```diff
@@ -454,6 +454,10 @@ pub struct CredSspServer<C: CredentialsProxy<AuthenticationData = AuthIdentity>>
     credentials_handle: Option<CredentialsBuffers>,
     ts_request_version: u32,
     context_config: Option<ServerMode>,
+    /// Set when the security context completed but the final SPNEGO token (accept-completed +
+    /// `mechListMIC`) still had to be sent to the client: the client cannot send `pubKeyAuth`
+    /// until it has verified that token, so `pubKeyAuth` arrives on the *next* leg.
+    awaiting_pub_key_auth: bool,
```
TheBestTvarynka, line 460: Honestly, I do not like this field 😅 It seems like a workaround due to the bad state machine.

I explored the code more deeply and thought about it for some time.

The problem is in the server-side CredSSP state machine. It expects the encrypted public key immediately after the internal protocol (NTLM) completes authorization. And that is wrong. It worked in the past, before the big SPNEGO refactoring.

Instead of adding a new `bool` flag, I recommend improving the state machine. I already developed a fix and will create a PR today.

Thank you for showing up to this problem!

TheBestTvarynka (commented):
Hi,
Thank you for the fix! 💟 And sorry for the late review. I was on a vacation last week 😅

I understand how we allowed this to happen. We have tests for `CredSSP + bare NTLM` and `CredSSP + SPNEGO + Kerberos`. But we did not have `CredSSP + SPNEGO + NTLM`.

> since Windows RDP servers reject bare NTLM inside CredSSP

I don't think it's true. As far as I know, the `mstsc` (default Windows RDP client) always sends bare NTLM over CredSSP, and it has always worked well.

I left a comment about your approach.

TheBestTvarynka:
closed in favor of #689

kihyun1998:
Yeah, this is a much better fix. The flag I added was really just a workaround, your state machine change actually fixes it. Thanks for taking the time to dig into it.

I'll send the `__test-data` CI matrix bit as its own PR so these client_server tests actually run in CI. They never have, which is probably why #687 went unnoticed for so long.

PR #689, opened 2026-06-16
fix(credssp): server side CresSSP + SPNEGO + NTLM

### LINKED ISSUE #687: CredSspServer (Negotiate/NTLM) cannot complete a handshake with sspi's own CredSspClient: pubKeyAuth expected one leg too early

TheBestTvarynka:
Hi,

This PR fixes the same problem as #688. I rework the fix. See this comment for more details: https://github.com/Devolutions/sspi-rs/pull/688#discussion_r3420080799

closes #687

### DIFF
--- a/src/credssp/mod.rs
+++ b/src/credssp/mod.rs
@@ -166,6 +166,7 @@ impl EarlyUserAuthResult {
 enum CredSspState {
     NegoToken,
     AuthInfo,
+    PubKeyInfo,
     Final,
 }
 
@@ -412,7 +413,7 @@ impl CredSspClient {
 
                 Ok(ClientState::FinalMessage(ts_request))
             }
-            CredSspState::Final => Err(Error::new(
+            CredSspState::Final | CredSspState::PubKeyInfo => Err(Error::new(
                 ErrorKind::OutOfSequence,
                 "CredSSP client's 'process' method must not be fired after the 'Finished' state",
             )),
@@ -484,6 +485,31 @@ impl<C: CredentialsProxy<AuthenticationData = AuthIdentity> + Send> CredSspServe
             ts_request_version,
             context_config: Some(client_mode),
         })
+    }
+
+    fn exchange_pub_key_auth(
+        &mut self,
+        pub_key_auth: &[u8],
+        client_nonce: &Option<[u8; NONCE_SIZE]>,
+    ) -> crate::Result<Vec<u8>> {
+        let peer_version = self
+            .context
+            .as_ref()
+            .unwrap()
+            .peer_version
+            .expect("an decrypt public key server function cannot be fired without any incoming TSRequest");
+
+        let context = self.context.as_mut().unwrap();
+
+        context.decrypt_public_key(
+            &self.public_key,
+            pub_key_auth,
+            EndpointType::Server,
+            client_nonce,
+            peer_version,
+        )?;
+
+        context.encrypt_public_key(&self.public_key, EndpointType::Server, client_nonce, peer_version)
     }
 
     #[instrument(fields(state = ?self.state), skip_all)]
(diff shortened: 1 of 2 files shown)


CBenoit (approved):
LGTM! Thank you for the initial research @kihyun1998

PR #691, opened 2026-06-18
ci: run the client_server test suite (enable __test-data in the matrix)

kihyun1998:
The `tests/sspi/client_server` module is gated on `all(network_client, __test-data)`, but no test-matrix row enables `__test-data` — so the client↔server pairing tests (`credssp_ntlm`, `credssp_kerberos`, and the `credssp_negotiate_ntlm` test merged in #689) never run in CI. That gap is how #687 shipped unnoticed.

**Commit 1 — CI:** add `__test-data` to the three root-manifest matrix rows (win/osx/linux).

**Commit 2 — test fix the flip surfaced:** with the suite finally compiling in CI, five `ServerProperties` initializers under `__test-data` were missing the `additional_service_keys` field (added to the struct at some point, but the never-compiled test code wasn't updated). Set to `Vec::new()` — the same default the library uses in `ServerProperties::new` / `fake_server_properties`; these tests use a single service, so behavior is unchanged. Kept as a separate commit; it's a prerequisite the CI change reveals, not unrelated work.

The suite is hermetic (`KdcMock` / `NetworkClientMock`, no real network). Verified across the full **win/osx/linux** matrix — all green.

Follow-up to #688 / #689.


TheBestTvarynka (approved):

TheBestTvarynka:
@kihyun1998, thank you!

CBenoit (approved):
Good catch, thank you!

PR #694, opened 2026-06-23
feat: support for Kerberos cross realm referral

thenextman:
RDP/NLA via FreeRDP and sspi-rs failed when a parent-domain user (richard@rjm.local) authenticated to a host in a child domain (TERMSRV/WIN-UE7FOENEK0D.dev.rjm.local) with

> InvalidToken: Asn1 error: "Expected Application number tag 15 but got: 30"

(Kerberos app tags: 15 = `AP-REP`, 30 = `KRB-ERROR` — i.e. we expected an `AP-REP` from the server but got a `KRB-ERROR`.)

### Root cause

sspi-rs has no cross-realm referral chasing. A KDC only issues tickets for principals in its own realm; for a service in another realm it returns a referral `TGT` (sname = krbtgt/<NEXT_REALM>), and the client must re-send the `TGS-REQ` to that next realm. Instead, sspi-rs:

1. Sent the `TGS-REQ` to the home KDC (RJM.LOCAL) → got back a referral `TGT` for krbtgt/DEV.RJM.LOCAL.
2. Stuffed that referral `TGT` directly into the `AP_REQ` as if it were the service ticket.
3. The target (WIN-UE7FOENEK0D) couldn't decrypt a ticket encrypted with the trust key; returned `KRB-ERROR` 41 (`KRB_AP_ERR_MODIFIED`).
4. sspi-rs tried to parse that `KRB-ERROR` as an `AP-REP` ---> `InvalidToken`.

### What we changed

Referral chasing loop in the `TGS` exchange (src/kerberos/client/mod.rs): after each `TGS-REP`, if the returned ticket sname is krbtgt/<NEXT_REALM>, re-issue the `TGS-REQ` for the same SPN to that realm using the referral `TGT`, chaining the session key and authenticator each hop, until the real service ticket comes back. Bounded to 10 hops with a no-progress guard.

Per-realm KDC routing (src/kerberos/mod.rs): split send into `send`/`send_to` and added `send_for_realm`. The pinned kdc_url (KDC proxy) stays authoritative for the home realm only; referral hops resolve through `detect_kdc_url(realm)`.

Helper and tests: extracted` referral_target_realm(sname)` (testable predicate) with 5 unit tests.

### Key decisions

Reuse the existing per-realm resolution chain rather than add new config. Referral hops route via the existing `SSPI_KDC_URL_<REALM>` env → krb5.conf → DNS SRV (_kerberos._tcp.<realm>) fallback. No new FFI/config plumbing; set one env var (or rely on DNS) and it works.

Pinned KDC = home realm only. A KDC proxy pinned to the home DC can't decrypt a krbtgt/<NEXT_REALM> referral ticket, so referral hops must reach the target realm's KDC.

### For further discussion

(Because I'm trying to keep these changes focussed)

Windows DNS SRV limitations ([src/dns.rs](vscode-webview://0o7221v7omgideiac71911ugm38hnasf8rt28rpk2ejj6sje9p7m/src/dns.rs)): the Windows path forces port :88 (ignores the SRV port) and reads only the first SRV record (no multi-DC failover).

Dotted env-var friction: SSPI_KDC_URL_DEV.RJM.LOCAL breaks PowerShell $env: parsing. Add underscore-normalized aliasing (SSPI_KDC_URL_DEV_RJM_LOCAL)?

U2U + referral interaction: `additional_tickets` is only carried on the first hop; edge case, likely fine, but worth a sanity check.

### DIFF
--- a/src/kerberos/client/mod.rs
+++ b/src/kerberos/client/mod.rs
@@ -10,7 +10,7 @@ pub use change_password::change_password;
 use picky_asn1_x509::oids;
 use picky_krb::constants::gss_api::{AP_REP_TOKEN_ID, AP_REQ_TOKEN_ID, AUTHENTICATOR_CHECKSUM_TYPE, TGT_REQ_TOKEN_ID};
 use picky_krb::crypto::CipherSuite;
-use picky_krb::data_types::{KrbResult, ResultExt};
+use picky_krb::data_types::{KrbResult, PrincipalName, ResultExt};
 use picky_krb::messages::{ApRep, TgsRep};
 use rand::rngs::{StdRng, SysRng};
 use rand_core::{Rng as _, SeedableRng as _};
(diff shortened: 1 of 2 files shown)


src/kerberos/client/mod.rs
```diff
@@ -276,34 +282,98 @@ pub async fn initialize_security_context<'a>(
...
-                    "the KDC reply message is too small: expected at least 4 bytes",
-                ));
-            }
-
-            // first 4 bytes are message len. skipping them
-            let mut d = picky_asn1_der::Deserializer::new_from_bytes(&response[4..]);
-            let tgs_rep: KrbResult<TgsRep> = KrbResult::deserialize(&mut d)?;
-            let tgs_rep = tgs_rep?;
+            // Cross-realm referral chasing (RFC 4120 §3.3.3.2 / MS-KILE).
+            //
```
TheBestTvarynka, line 286: nice comment 👍 

I would also add direct links to the documentation:


```suggestion
            // Cross-realm referral chasing (RFC 4120 §3.3.3.2 / MS-KILE).
            //
            // * [Cross-Realm Operation](https://www.rfc-editor.org/rfc/rfc4120.html#section-1.2)
            // * [Server Referrals](https://www.rfc-editor.org/rfc/rfc6806.html#section-8)
```

The second link is especially useful because that section even contains an example of the mentioned hops

TheBestTvarynka (approved):
LGTM

PR #698, opened 2026-06-24
feat(kerberos): honour DNS SRV port and fail over across multiple KDCs

thenextman:
ℹ️  1. This doesn't come from any specific client issue or concern; it's more of a feature gap that I'd like to close before it becomes an issue and a personal itch to scratch.

ℹ️ 2. Primarily engineered with Claude and dev testing still to be completed.

So, at this point the PR is more to validate the correctness of the approach, underlying code, etc

-------

### Issues

- DNS SRV port ignored on Windows. `detect_kdc_hosts_from_dns_windows` hardcoded :88 and read only the first SRV record
- Multiple KDCs never used. Resolution collapsed to the first candidate (`detect_kdc_host().first()`), and the send path tried one KDC with no failover if it was down.
- krb5.conf returned only the first kdc = line. Additional KDCs for a realm were silently dropped.
- No KDC stickiness across an auth. Each Kerberos message (AS ×2, TGS, referral hops) re-resolved independently, so it could bounce between KDCs under round-robin DNS.

### Changes

1. Honour the SRV port everywhere. Windows now walks the `DnsQuery_W` results, filters to SRV records and uses the record's `wPort`. 
2. Resolve and fail over across multiple KDCs. New `detect_kdc_urls()` returns the full ordered candidate list (DNS SRV / Windows KdcNames / krb5.conf); `send_for_realm` tries each in order. `detect_kdc_url()` kept (first candidate) for backward compatibility.
3. krb5.conf multi-KDC. `Krb5Conf::get_all_values` returns every kdc = entry; one-host-per-line.
4. Per-realm KDC pinning with fallback. The KDC that answers for a realm is cached and reused for the rest of the exchange, re-resolved only if it later fails. Failover advances only on a transport failure; any KDC reply, including a `KRB-ERROR`, is returned as-is.

### Decisions

1. krb5.conf = one host per kdc = line with no whitespace splitting. I confirmed this matches the MIT documentation; although in the real world we saw at least one case of multiple KDCs on a single line I think it's better to follow the officially documented format here.
2. Failover timeout keeps the existing 5s per-attempt (`KDC_CONNECT_TIMEOUT`).
3. Stickiness pins on first success and then falls back on failure. Not a hard pin and does not re-resolve every message (see below).
4. `&mut self` and plain `HashMap`, not `RefCell`: the generator's futures are + `Send`, so a `RefCell` (`!Sync`) would break the bound. `&mut` self threaded through all callers without borrow conflicts.
5. Referral routing unchanged. `self.realm` stays the home realm; the per-realm cache is keyed independently, so the home-vs-child KDC decision is preserved.

-------

For pinning a KDC across requests, I was unsure what to do here. As I understand it, a message sequence is not tied to a specific KDC i.e. requests in a single exchange can round-robin between different KDCs. It seems an inherent part of the design. However I think we should avoid the lookup on every request if possible.

MIT does a staggered, parallel request (i.e. it tries _all_ KDCs with a stagger and backoff). There's no pinning after successfully talking to a KDC, although the used KDC is passed back to the client in an `out` parameter so the client can pin if they desire to. This seems an optimal solution but also has significant complexity.

In Windows, as far as I can tell, this is handled in `netlogon` rather than the Kerberos client. The [documentation](https://learn.microsoft.com/windows-server/identity/ad-ds/manage/dc-locator) says:

- "The Netlogon service caches the discovered domain controller… Caching this information encourages the consistent use of the same domain controller."
- It returns "the information… from the domain controller that responds first" (a responsiveness-based pick), then caches it.
- Negative caching exists too (the shutdown-DC KB: a client can land in "negative cache mode").

So it seems Windows supports the spirit of "pinning" to a DC, but it's the OS doing that across the whole machine, not a per-auth context cache on the Kerberos side.

I consider what we have here a reasonable middle-ground; not a faithful copy of MIT or Microsoft but its own thing. There's some complexity overhead, if we wanted to simply drop the pinning and re-resolve on every request it simplifies the code somewhat.

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -11,7 +11,11 @@ cfg_if::cfg_if! {
         use std::ptr::{null_mut};
         use core::ffi::{c_void};
 
-        pub(crate) fn dns_query_srv_records(name: &str) -> Vec<String> {
+        /// Queries SRV records for `name`, returning each `(target, port)`.
+        ///
+        /// `DnsQuery_W` yields a linked list that may include non-SRV records (e.g. A records for
+        /// the SRV targets), so we walk `pNext` and keep only SRV entries.
+        pub(crate) fn dns_query_srv_records(name: &str) -> Vec<(String, u16)> {
             let mut records = Vec::new();
 
             let mut p_query_results: *mut DNS_RECORDA = null_mut();
@@ -30,23 +34,29 @@ cfg_if::cfg_if! {
 
             match dns_status.ok() {
                 Ok(()) => {
-                    // SAFETY: `p_query_results` in non-null because `dns_status` is Ok.
-                    let query_results = unsafe { *p_query_results };
-                    // SAFETY: `query_results.Data` is guaranteed to contain `SRV` because
-                    // of arguments to `DnsQuery_W`.
-                    let p_name_target = unsafe { query_results.Data.Srv.pNameTarget };
-                    let name_target = PWSTR::from_raw(p_name_target.as_ptr() as *mut u16);
-                    // SAFETY: `name_target` is guaranteed to be correct at this point.
-                    let name_target = unsafe {name_target.to_string()};
+                    let mut p_record = p_query_results;
+                    while !p_record.is_null() {
+                        // SAFETY: `p_record` is non-null and points to a valid `DNS_RECORDA`.
+                        let record = unsafe { *p_record };
 
-                    if let Ok(name_target) = name_target {
-                        records.push(name_target);
+                        if record.wType == DNS_TYPE_SRV.0 {
+                            // SAFETY: `wType == DNS_TYPE_SRV` guarantees the `Srv` union member is active.
+                            let srv = unsafe { record.Data.Srv };
+                            // `DnsQuery_W` returns wide strings even in the `*A` record struct.
+                            let name_target = PWSTR::from_raw(srv.pNameTarget.as_ptr() as *mut u16);
+                            // SAFETY: `name_target` points to a valid, null-terminated UTF-16 string.
+                            if let Ok(name_target) = unsafe { name_target.to_string() } {
+                                records.push((name_target, srv.wPort));
+                            }
+                        }
+
+                        p_record = record.pNext;
                     }
                 }
                 Err(error) => error!(%error, "DnsQuery_W failed"),
             }
 
-            // SAFETY: `p_query_results` is not null.
+            // SAFETY: `DnsFree` tolerates a null pointer.
             unsafe {
                 DnsFree(Some(p_query_results as *const c_void), DnsFreeRecordList);
             }
@@ -136,14 +146,14 @@ cfg_if::cfg_if! {
             let krb_tcp_srv = dns_query_srv_records(krb_tcp_name);
 
             if !krb_tcp_srv.is_empty() {
-                return krb_tcp_srv.iter().map(|x| format!("tcp://{x}:88")).collect()
+                return srv_records_to_kdc_urls("tcp", &krb_tcp_srv)
             }
 
             let krb_udp_name = &format!("_kerberos._udp.{domain}");
             let krb_udp_srv = dns_query_srv_records(krb_udp_name);
 
             if !krb_udp_srv.is_empty() {
-                return krb_udp_srv.iter().map(|x| format!("udp://{x}:88")).collect()
+                return srv_records_to_kdc_urls("udp", &krb_udp_srv)
             }
 
             Vec::new()
@@ -334,14 +344,16 @@ cfg_if::cfg_if! {
             let krb_tcp_srv = dns_query_srv_records(krb_tcp_name);
 
             if !krb_tcp_srv.is_empty() {
-                return krb_tcp_srv.iter().map(|x| format!("tcp://{}:{}", &x.target, x.port)).collect()
+                let records: Vec<(String, u16)> = krb_tcp_srv.into_iter().map(|x| (x.target, x.port)).collect();
+                return srv_records_to_kdc_urls("tcp", &records)
             }
 
             let krb_udp_name = &format!("_kerberos._udp.{domain}");
             let krb_udp_srv = dns_query_srv_records(krb_udp_name);
 
             if !krb_udp_srv.is_empty() {
-                return krb_udp_srv.iter().map(|x| format!("udp://{}:{}", &x.target, x.port)).collect()
+                let records: Vec<(String, u16)> = krb_udp_srv.into_iter().map(|x| (x.target, x.port)).collect();
+                return srv_records_to_kdc_urls("udp", &records)
             }
 
             Vec::new()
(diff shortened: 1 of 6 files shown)


src/dns.rs
```diff
@@ -397,6 +401,17 @@ where
...
 
+/// Builds `scheme://host:port` KDC URLs from resolved SRV `(target, port)` records.
+///
+/// DNS returns absolute names with a trailing dot (`dc.example.com.`); it is stripped so the
+/// result parses as a URL host.
+fn srv_records_to_kdc_urls(scheme: &str, records: &[(String, u16)]) -> Vec<String> {
+    records
+        .iter()
+        .map(|(target, port)| format!("{scheme}://{}:{port}", target.trim_end_matches('.')))
+        .collect()
```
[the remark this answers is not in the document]
thenextman: Documented as a known limitation, not for this PR.

src/kerberos/mod.rs
```diff
@@ -166,33 +172,80 @@ impl Kerberos {
...
+                        %error,
+                        candidate = index + 1,
+                        candidate_count,
+                        realm,
+                        "KDC request failed; trying next candidate if available"
+                    );
+                    last_error = Some(error);
+                }
+            }
+        }
```
CBenoit, line 241: thought: I think this loop is not covered, that’s where the whole orchestration takes place (sticky-cache hit/reuse, the already_tried skip, error accumulation, cache insert-on-success / remove-on-failure…). Since it’s the most behaviorally complex part of the PR it would be nice to add a generator-harness test with a fake network that fails candidate 1 and succeeds on 2 (asserting the cache ends pinned to 2, and that a later failure of 2 re-resolves).

src/kerberos/mod.rs
```diff
@@ -166,33 +172,80 @@ impl Kerberos {
...
-    /// krb5.conf / DNS SRV records) are honored for the referral hop. This lets a single auth
-    /// chase a referral into a child/trusted realm without changing the pinned home-realm KDC.
-    async fn send_for_realm(&self, yield_point: &mut YieldPointLocal, realm: &str, data: &[u8]) -> Result<Vec<u8>> {
-        let kdc_url = if self.realm.as_deref() == Some(realm) {
-            self.kdc_url.clone().or_else(|| detect_kdc_url(realm))
-        } else {
-            detect_kdc_url(realm)
+    /// The KDC that last answered for `realm` is reused across the messages of an auth; it is only
+    /// re-resolved if it stops responding. When resolving, candidates are tried in order. Failover
+    /// (sticky or in-order) advances only on a *transport* failure (could not connect / no reply);
```
CBenoit, line 187: suggestion: "on any error" may be more accurate. `send_to` below also returns Err for a bad URL scheme or an under-length UDP message, etc. None are transport failures. Harmless in practice: scheme/UDP-length errors are deterministic across candidates from one detection (same scheme, same bytes), so the loop just ends on the last error, and failing over from a misbehaving proxy is arguably desirable.

src/dns.rs
```diff
@@ -397,6 +401,21 @@ where
...
 }
 
+/// Builds `scheme://host:port` KDC URLs from resolved SRV `(target, port)` records.
+///
+/// DNS returns absolute names with a trailing dot (`dc.example.com.`); it is stripped so the
+/// result parses as a URL host.
+///
+/// Limitation: SRV priority/weight (RFC 2782) are not applied here — candidates keep the order the
+/// platform resolver returned (Windows `DnsQuery_W` already sorts by priority/weight; hickory and
+/// `async_dnssd` do not). Proper priority ordering + weighted selection is a follow-up.
```
CBenoit, line 506: question: Should we open an issue?

src/kdc.rs
```diff
@@ -80,8 +93,19 @@ pub fn detect_kdc_host(domain: &str) -> Option<String> {
 }
 
 pub fn detect_kdc_url(domain: &str) -> Option<Url> {
-    let kdc_host = detect_kdc_host(domain)?;
-    Url::from_str(&kdc_host).ok()
+    detect_kdc_urls(domain).into_iter().next()
+}
```
CBenoit, line 97: note: Looks like an improvement over the previous code which was returning None on the first KDC host even if unparseable.

src/krb.rs
```diff
@@ -85,6 +85,20 @@ impl Krb5Conf {
...
+    ///
+    /// Unlike [`get_value`](Self::get_value), which returns only the first match, this preserves
+    /// repeated keys — e.g. the multiple `kdc = ` lines (one host each) a realm uses to list
+    /// several KDCs for failover.
+    pub(crate) fn get_all_values(&self, path: Vec<&str>) -> Vec<String> {
+        let path = path.join("|");
+        self.values
+            .iter()
+            .filter(|(key, _)| key.eq_ignore_ascii_case(&path))
+            .map(|(_, val)| val.clone())
```
CBenoit, line 98: nitpick: This clones every matched value. Not a big deal for a tiny, cold config path; but I think it’s trivial to return the borrowed `&str` instead. The callsite doesn’t really need owned strings from what I can tell.

src/dns.rs
```diff
@@ -30,23 +34,29 @@ cfg_if::cfg_if! {
...
+                    let mut p_record = p_query_results;
+                    while !p_record.is_null() {
+                        // SAFETY: `p_record` is non-null and points to a valid `DNS_RECORDA`.
+                        let record = unsafe { *p_record };
+
+                        if record.wType == DNS_TYPE_SRV.0 {
+                            // SAFETY: `wType == DNS_TYPE_SRV` guarantees the `Srv` union member is active.
+                            let srv = unsafe { record.Data.Srv };
+                            // `DnsQuery_W` returns wide strings even in the `*A` record struct.
+                            let name_target = PWSTR::from_raw(srv.pNameTarget.as_ptr() as *mut u16);
```
CBenoit, line 46: suggestion: `DNS_RECORDA` + wide-string cast is pre-existing but now it looks slightly more risky.

The current code looks sound, but this cast is now the mechanism for multi-KDC discovery on Windows, so it carries a bit more weight than before.

The code can be made less confusing by confining the reinterpret to one upfront cast and let the types tell the truth (via `DNS_RECORDW`) for the rest of the loop:

```rust
    // DnsQuery_W returns wide strings; the binding types the list as DNS_RECORDA, so reinterpret
    // the whole list as DNS_RECORDW once. Layouts are identical (PSTR/PWSTR are both pointers).
    let mut p_record = p_query_results.cast::<DNS_RECORDW>();
    while !p_record.is_null() {
        let record = unsafe { *p_record };
        if record.wType == DNS_TYPE_SRV.0 {
            // pNameTarget is now a PWSTR — no per-record cast.
            let srv = unsafe { record.Data.Srv };
            if let Ok(name_target) = unsafe { srv.pNameTarget.to_string() } {
                records.push((name_target, srv.wPort));
            }
        }
        p_record = record.pNext; // *mut DNS_RECORDW — walk stays wide
    }
    unsafe { DnsFree(Some(p_query_results as *const c_void), DnsFreeRecordList) };
```

Same single cast, but moved from inside the loop to one spot at the boundary. At this point, `.to_string()` is both the correct and obvious method to call. `DnsFree` is encoding-agnostic.

CBenoit (commented):
Thank you, the code looks good to me. I’ve found a few places that may need some attention before we merge, see below.

TheBestTvarynka (approved):

PR #706, opened 2026-07-15
fix(dns): avoid panic on short SRV record RDATA on macOS

thenextman:
We've received crash reports from a client application on macOS showing

> thread " (46265) panicked at src/dns.rs:172:56: range end index 2 out of range for slice of length O
note: run with 'RUST_BACKTRACE=1' environment variable to display a backtrace

It's the first we've seen of it, but this client reports that it crashes _consistently_ in his environment.

The cited line lands exactly on `rdata[0..2]` inside the SRV-record parser. Presumably the `rdata` buffer was empty so slicing it caused a panic. But why is the record empty? I'm not sure.

Every callback from the DNS query returns a `DnsSrvRecord `, where the RDATA is at least 6 bytes (priority + weight + port, then the target name). The code assumes a well formed SRV record.  We see some (potential) defects in the code:

1. The parser is unguarded and trusts `rdata` to always be at least 6 bytes
2. I believe (but am not sure) that the normal "this realm has no SRV record" outcome is **not** an empty callback, but a timeout. mDNSResponder doesn't call back in time, our timeout fires and we return an empty list. I guess in this case _something_ causes the SRV query to be answered immediately, with a negative (empty) response. I'm equally not sure what that could be: VPN resolver, enterprise DNS, a local stub resolver.... Not sure.

The (blind) fix:

In `src/dns.rs`, in the macOS/iOS path:

- From → TryFrom, returning `Err` when `rdata.len() < 6` instead of slicing
- The query loop now filters on the `ADD` flag (the correct signal for "this is a real, added record"), skipping negative responses and removals rather than parsing them.
- Both skip cases are logged (`debug!`/`warn!`) with the record's `rr_type`, `rr_class`, and `rdata_len`, so if a resolver sends something unexpected again, we can see exactly what it returned instead of guessing.

For a realm with no SRV records, the empty/negative answer is ignored, we now fall through to the UDP query and then return an empty host list.

### DIFF
--- a/src/dns.rs
+++ b/src/dns.rs
@@ -153,6 +153,7 @@ cfg_if::cfg_if! {
 
 cfg_if::cfg_if! {
     if #[cfg(any(target_os="macos", target_os="ios"))] {
+        use std::fmt;
         use std::time::Duration;
         use tokio::time::timeout;
         use futures::stream::{StreamExt};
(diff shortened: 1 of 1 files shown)


src/dns.rs
```diff
@@ -166,19 +166,29 @@ cfg_if::cfg_if! {
...
+            fn try_from(record: &QueryRecordResult) -> Result<Self, Self::Error> {
                 let rdata = record.rdata.as_slice();
+                // The minimum size of an SRV record's RDATA is 6 bytes: 2 (priority)
+                // + 2 (weight) + 2 (port), followed by the target name. Anything
+                // shorter cannot be a valid SRV record and must not be indexed into.
+                // In practice this only happens for the negative/removal callbacks
+                // handled by the caller, so this is a defensive guard.
+                if rdata.len() < 6 {
+                    return Err(());
+                }
```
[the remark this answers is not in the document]
thenextman: Addressed by validating the decoded target rather than the byte length, since that also handles the RFC 2782 root-target case

src/dns.rs
```diff
@@ -208,8 +226,37 @@ cfg_if::cfg_if! {
...
+                                    Err(()) => {
+                                        warn!(
+                                            name,
+                                            rr_type = ?dns_record.rr_type,
+                                            rr_class = ?dns_record.rr_class,
+                                            rdata_len = dns_record.rdata.len(),
+                                            "Ignoring DNS SRV callback with RDATA too short to be an SRV record"
+                                        );
+                                    }
+                                }
```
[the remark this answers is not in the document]
thenextman: The `try_from` error is now a `&'static str` reason ("RDATA shorter than the 6-byte SRV fixed fields" or "SRV target names no host"), and the loop logs it via reason under a broader "Ignoring malformed DNS SRV callback" message

src/dns.rs
```diff
@@ -166,19 +166,38 @@ cfg_if::cfg_if! {
             target: String
         }
 
-        impl From<&QueryRecordResult> for DnsSrvRecord {
-            fn from(record: &QueryRecordResult) -> Self {
+        impl TryFrom<&QueryRecordResult> for DnsSrvRecord {
+            /// Why an SRV callback could not be turned into a usable record, for diagnostics.
+            type Error = &'static str;
```
CBenoit: nitpick: There are only two very concrete error cases, so I would use a small error type implementing the Error trait, as an enum:
```rust
#[derive(Debug)]
enum TryFromQueryRecordResultError {
    RdataTooShort,
    EmptySrvTarget,
}

// + implement Display with basically the strings already defined
```
(name is bikeshedable)

It’s a bit of boilerplate, but it works better than `&'static str` thanks of the Error trait.
Currently it’s not relevant though, as the only consumer just print this as a reason string.

src/dns.rs
```diff
@@ -208,8 +227,38 @@ cfg_if::cfg_if! {
...
+                                            name,
+                                            reason,
+                                            rr_type = ?dns_record.rr_type,
+                                            rr_class = ?dns_record.rr_class,
+                                            rdata_len = dns_record.rdata.len(),
+                                            "Ignoring malformed DNS SRV callback"
+                                        );
+                                    }
+                                }
+                            }
```
CBenoit, line 307: Claude flagged this:

> Filtering on ADD diverges slightly from RFC 2782's "single root-target SRV = abort" rule, but strictly improves on the prior behavior
dns.rs:286-302 (detect_kdc_hosts_from_dns_apple): RFC 2782 §"Usage rules" says that if the only SRV RR returned for a query is Target = ".", the client should abort rather than try alternate protocols/records. This PR's effect is that such a record is filtered out (empty target) and the code falls through to the UDP query instead of aborting. This is a pre-existing characteristic of this code path (the old code would have pushed a hostless URL rather than aborting either), so it isn't a regression introduced here, and the new behavior is objectively safer than emitting "tcp://:88". Flagging only because a stricter RFC-2782 read would treat this as "authoritatively no service" rather than "try the next transport" — likely not worth chasing for a hotfix, but worth knowing about if KDC discovery correctness in RFC-2782-compliant environments ever comes up again.

Maybe a follow up? cc @PavloMyroniuk-apriorit

CBenoit (approved):
Thank you, LGTM!

I have a few suggestions / nitpicks if you want to address before merging.

One nitpick is that we could have a small regression test; although I don’t really see that regressing in the future tbh.

thenextman:
@CBenoit I addressed your comments but, I'm sorry, Claude force pushed an amend instead of a new commit before I could stop him 😢

src/dns.rs
thenextman: It's true, but yes I think it could be a follow up. I think the proper semantics aren't worth it for what is really a hot fix.

Issue #708, opened 2026-07-15
Remove `Username::domain_name()` in favor of `Username::parts()`


CBenoit:
### Background

`Username::domain_name()` returns `Option<&str>` but the `&str` means two different things depending on the internal format:

- **UPN** (`user@corp.example.com`) → returns the UPN **suffix** (`corp.example.com`)
- **Down-level logon name** (`CORP\user`) → returns the **NetBIOS domain** (`CORP`)

A UPN suffix and a NetBIOS domain are not interchangeable, but the accessor hands them back through the same type with no signal of which one you got. A caller that wants "the NetBIOS domain for a down-level logon" and calls `domain_name()` on a UPN silently receives a DNS-style suffix instead, the wrong value, shaped plausibly enough to flow downstream undetected.

#709 adds a format-tagged replacement, `Username::parts() -> UsernameParts<'_>`, whose UPN arm exposes only `suffix` and whose down-level arm exposes only `netbios_domain: Option<&str>`. Because the enum is matched exhaustively, treating a UPN suffix as a NetBIOS domain no longer compiles. `domain_name()` is deprecated in that PR; this issue tracks its removal, which is a breaking change and must ship in its own `fix!`/`feat!` release.

### Why this is a real footgun, not a theoretical one

`impl From<AuthIdentity> for AuthIdentityBuffers` writes `domain_name().unwrap_or_default()` straight into the `domain` field of `AuthIdentityBuffers`, a slot consumed as a NetBIOS domain by the NTLM/Kerberos paths. For a UPN identity this places the DNS suffix where a NetBIOS domain is expected. This is the exact confusion the type-level split is meant to make impossible.

### External impact

`Username` is part of the published `sspi` crate's public API and is used by downstream consumers (IronRDP, Devolutions Gateway, and third parties). Removal is a breaking change:

### Checklist

- [ ] Land the deprecation (companion PR) in a minor release first, giving downstreams a compiler-visible migration window.
- [ ] Remove in a later dedicated breaking release, commit marked `fix!`/`feat!` with a `BREAKING CHANGE:` footer.

### Migration example for downstreams

```rust
// Before
let domain = username.domain_name().unwrap_or_default();

// After
let domain = match username.parts() {
    UsernameParts::DownLevelLogonName { netbios_domain, .. } => netbios_domain.unwrap_or_default(),
    UsernameParts::UserPrincipalName { .. } => "", // a UPN has no NetBIOS domain
};
```

PR #709, opened 2026-07-15
feat(sspi): add format-tagged Username::parts view

CBenoit:
Add `UsernameParts`, a format-tagged, borrowed view into a `Username`, and `Username::parts()` returning it. Each variant only exposes the fields that are meaningful for its user name format, so a UPN suffix can no longer be mistaken for a NetBIOS domain: the `UserPrincipalName` arm has a `suffix` (and the full `upn`) but no "domain" field, while `DownLevelLogonName` carries an `Option<netbios_domain>` that models accurately the no-separator case. Matching is exhaustive over the two Microsoft user name formats.

`Username::domain_name()` conflates a UPN suffix with a NetBIOS domain and is deprecated in favor of `Username::parts()`. Deprecation alone is non-breaking, so this is a minor bump. Removal is a future breaking release tracked by #708.

This also fixes three UPN-vs-NetBIOS confusion bugs, all rooted in `domain_name()` returning the UPN suffix and the NetBIOS domain indistinguishably:

1. Server credential matching matched across formats: `Negotiate::set_auth_identity` collapsed both formats to a single "domain", so UPN `user@EXAMPLE` and DLLN `EXAMPLE\user` compared equal even though they are distinct identities.

2. UPN format lost through `AuthIdentityBuffers` round-trip: UPNs were stored as `(user=account, domain=suffix)`, but the reverse conversion reads `domain` as a NetBIOS domain, degrading `alice@example.com` into `example.com\alice`.

3. Realm split on the wrong `@`: `get_client_principal_realm_impl` used the first `@` while `Username::parse` uses the last, so `user@dept@example.com` derived the wrong realm.

The new format-tagged `Username::parts()` API makes this class of confusion unrepresentable.

### DIFF
--- a/src/auth_identity.rs
+++ b/src/auth_identity.rs
@@ -384,13 +384,20 @@ impl From<AuthIdentity> for AuthIdentityBuffers {
     fn from(credentials: AuthIdentity) -> Self {
         let password: &str = credentials.password.as_ref().as_ref();
 
-        let domain = match credentials.username.parts() {
-            UsernameParts::UserPrincipalName { suffix, .. } => suffix,
-            UsernameParts::DownLevelLogonName { netbios_domain, .. } => netbios_domain.unwrap_or_default(),
+        // Encode the username so that it round-trips back to the same format through
+        // `TryFrom<&AuthIdentityBuffers>` (which parses `user` and treats `domain` as a NetBIOS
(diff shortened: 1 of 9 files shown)


TheBestTvarynka (approved):
🚀

PR #734, opened 2026-08-25
fix(sspi): RD Gateway auth

### LINKED ISSUE #729: SPNEGO NTLM authorization error when connecting to RD Gateway
Set up RD Gateway and try to connect via `FreeRDP` (do not worry about passwords; it's a local lab):

```bash
./sdl-freerdp /v:DESKTOP-QELPR32.tbt.com /gateway:g:dgateway.tbt.com,u:t2,d:TBT,p:qqqQQQ111!!! /u:t2 /d:TBT /cert:ignore /p:qqqQQQ111!!! /log-level:TRACE /sspi-module:libsspi.dylib /auth-pkg-list:ntlm,!kerberos /tls:secrets-file:tls-secrets.txt
```

The authorization process will fail:

Image

I already did a small investigation. The problem in the `SecurityStatus` returned to FreeRDP by sspi-rs. FreeRDP expects `SecurityStatus::Ok` but the library returns `SecurityStatus::ContinueNeeded`.

TheBestTvarynka:
closes #729 

Hi,
I fixed the RD Gateway auth in this PR.

# What's wrong with RD Gateway?

It does not send a confirmation (finalization) message back. When we send a Kerberos AP_REQ message wrapped in an SPNEGO token, the RD Gateway server doesn't send an AP_REP back. Moreover, when the last message from the client contains SPNEGO ACCEPT_INCOMPLETE, the server does not respond with SPNEGO ACCEPT_COMPLETE. Instead, it processes further with a WebSocket upgrade:

image

Even when the SPNEGO MIC token exchange is needed, the RD Gateway server still doesn't send the MIC token back and continues with the WebSocket upgrade.

# Why does FreeRDP fail to connect?

Because `sspi-rs` waits for confirmation and returns `SecurityStatus::ContinueNeeded`. FreeRDP doesn't expect this status code and interrupts auth. FreeRDP expects the last NLA status code to be `SecurityStatus::Ok` before processing the WebSocket upgrade.

To fix this, we need to recognize RD Gateway auth and return `SecurityStatus::Ok` immediately.

# The fix

The fix is simple: `sspi-rs` detects RD Gateway auth by checking the target name. If it starts with "HTTP/", then we return `SecurityStatus::Ok` without waiting for the confirmation/response.

The fix may not feel right, but there is no other good way to do it (at least for now). Ideally, I would like to use the `ClientRequestFlags::MUTUAL_AUTH` flag here. If it is not present, we return `SecurityStatus::Ok` without waiting for the response. But FreeRDP provides this flag in all cases, so it is not an option.

# Native SSPI behavior

I was curious about how it works in native SSPI. Turned out that native SSPI returns `SecurityStatus::ContinueNeeded` (as we did). `mstsc` ignores it and continues with the WebSocket upgrade. The `ClientRequestFlags::MUTUAL_AUTH` flag is also present.

In other words, this fix is caused by the behavior difference between FreeRDP and `mstsc` in the RD Gateway implementation.

# Testing

I added a few tests to cover this case. I also dev-tested auth using FreeRDP (both NTLM and Kerberos protocols). Users should be able to connect via RD Gateway using NTLM or Kerberos when this PR is merged.

### DIFF
--- a/src/kerberos/client/mod.rs
+++ b/src/kerberos/client/mod.rs
@@ -463,7 +463,7 @@ pub async fn initialize_security_context<'a>(
 
             // "HTTP/" prefix of the target name means RD Gateway authorization.
             // The RD Gateway server does not send an AP_REP message, so we skip the AP_REP processing and go directly to the final state.
-            let security_status = if let Some(target_name) = &builder.target_name
+            if let Some(target_name) = &builder.target_name
                 && target_name.starts_with("HTTP/")
             {
                 client.state = KerberosState::Final;
@@ -473,9 +473,7 @@ pub async fn initialize_security_context<'a>(
                 client.state = KerberosState::ApExchange;
 
                 SecurityStatus::ContinueNeeded
-            };
-
-            security_status
+            }
         }
         KerberosState::ApExchange => {
             let input = builder
--- a/src/kerberos/client/mod.rs
+++ b/src/kerberos/client/mod.rs
@@ -461,9 +461,21 @@ pub async fn initialize_security_context<'a>(
             let output_token = SecurityBuffer::find_buffer_mut(builder.output, BufferType::Token)?;
             output_token.buffer = encoded_neg_ap_req;
 
-            client.state = KerberosState::ApExchange;
+            // "HTTP/" prefix of the target name means RD Gateway authorization.
+            // The RD Gateway server does not send an AP_REP message, so we skip the AP_REP processing and go directly to the final state.
+            let security_status = if let Some(target_name) = &builder.target_name
+                && target_name.starts_with("HTTP/")
+            {
+                client.state = KerberosState::Final;
 
-            SecurityStatus::ContinueNeeded
+                SecurityStatus::Ok
+            } else {
+                client.state = KerberosState::ApExchange;
+
+                SecurityStatus::ContinueNeeded
+            };
+
+            security_status
         }
         KerberosState::ApExchange => {
             let input = builder
--- a/src/negotiate/client.rs
+++ b/src/negotiate/client.rs
@@ -103,7 +103,7 @@ pub(crate) async fn initialize_security_context<'a>(
                 .initialize_security_context(negotiate.auth_identity.as_ref(), yield_point, builder)
                 .await;
 
-            let first_token = match result {
+            let (first_token, security_status) = match result {
                 Ok(result) => {
                     if result.status != SecurityStatus::ContinueNeeded && result.status != SecurityStatus::Ok {
                         return Err(Error::new(
@@ -113,7 +113,7 @@ pub(crate) async fn initialize_security_context<'a>(
                     }
 
                     let token = SecurityBuffer::find_buffer_mut(builder.output, BufferType::Token)?;
-                    Some(mem::take(&mut token.buffer))
+                    (Some(mem::take(&mut token.buffer)), result.status)
                 }
                 Err(err)
                     if matches!(negotiate.protocol, NegotiatedProtocol::Kerberos(_))
@@ -143,7 +143,7 @@ pub(crate) async fn initialize_security_context<'a>(
 
                     let token = SecurityBuffer::find_buffer_mut(builder.output, BufferType::Token)?;
 
-                    Some(mem::take(&mut token.buffer))
+                    (Some(mem::take(&mut token.buffer)), result.status)
                 }
                 Err(err) => {
                     return Err(err);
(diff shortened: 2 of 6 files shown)


thenextman:
Thank you @TheBestTvarynka!

As I understand it, we're doing the detection based on the web socket? My concern is, that we break things in an unexpected way when using RPC instead of web sockets. That might be why I could not reproduce the problem on my side (our lab environment could very well be configured for RPC instead of web sockets). Maybe my concern is unfounded because I don't understand the technical details :)

TheBestTvarynka:
> As I understand it, we're doing the detection based on the web socket?

Yes, you are right

> we break things in an unexpected way when using RPC instead of web sockets

I am not sure about that, but does the target name start with "HTTP/" when connecting over RPC? I thought it was only a case for WebSocket-based connections. Give me a few minutes. I will research it

TheBestTvarynka:
@thenextman Good news, it still works when connecting to the RD gateway over RPC instead of WebSocket.

I turned my lab on and added `type:rpc` to FreeRDP's `/gateway:` argument. As a result, a connection is successful:

image

When I looked at Wireshark, I think I saw a similar thing to a WebSocket-based connection:

image

But I want to mention that I could only test with plain NTLM (even without SPNEGO). Somewhat, when the RPC type is on, FreeRDP explicitly uses the "NTLM" package name instead of "Negotiate":

image

So, I am unable to try Kerberos for RPC-based RD Gateway auth. At least plain NTLM works.

thenextman:
Appreciate the investigation and it matches my experience. This is great for me, once @CBenoit take a look at the code changes (not urgent)

CBenoit (approved):
Thank you for the investigation! I was a little bit surprised by the heuristic, but I’ll trust your intuition about it. The code is looking good to me otherwise!

thenextman:
@TheBestTvarynka If we were to pursue a change on the FreeRDP side, do you think it would be better?

> Ideally, I would like to use the ClientRequestFlags::MUTUAL_AUTH flag here. If it is not present, we return  SecurityStatus::Ok without waiting for the response. But FreeRDP provides this flag in all cases, so it is not an option.

> in native SSPI [...] The ClientRequestFlags::MUTUAL_AUTH flag is also present.

It's a bit unclear to me; you said FreeRDP always provides this flag and mstsc also has it present? So where's the difference?

CBenoit:
I would prefer a proper fix through FreeRDP too, so we don’t need the "HTTP/" match heuristic, because it has a bigger breadth than ideally. Not against merging the PR if that’s the best thing we can do, but if we don’t need to rush I would wait a little bit more.

TheBestTvarynka:
> So where's the difference?

@thenextman It was just a suggestion. On FreeRDP's side, it is easier to remove one flag in the `InitializeSecurityContext` call than rewrite the whole RD Gateway authentication flow. We could use the absence of the `MUTUAL_AUTH` flag as a sign to return `SecurityStatus::Ok`. This would let us handle this case on our side without rewriting much of FreeRDP.

If we still decide to refactor FreeRDP, we need to change its behavior so it continues establishing the RD Gateway connection after receiving a successful HTTP code, even when the security package returned `SecurityStatus::ContinueNeeded` (`SEC_E_CONTINUE_NEEDED`) (`mstsc` does this)

TheBestTvarynka:
> If we were to pursue a change on the FreeRDP side, do you think it would be better?

@thenextman, I think yes, it would be better. Because checking if the target name starts with "HTTP/" is a workaround. Ideally, it should not be present in the code, but currently it is the only way to make it work with RD Gateway

CBenoit (changes requested):
I’m removing my approval for now, so it does not appear as a ready-to-merge PR

TheBestTvarynka:
@thenextman I tested the RD Gateway auth with your FreeRDP fix (https://github.com/FreeRDP/FreeRDP/pull/13418). Everything works well, so I can close this PR.

cc @CBenoit 

## RD Gateway logon (Kerberos)

```bash
./sdl-freerdp /v:DESKTOP-QELPR32.tbt.com /gateway:g:dgateway.tbt.com,u:t2,d:TBT,p:qqqQQQ111!!! /u:t2 /d:TBT /cert:ignore /p:qqqQQQ111!!! /log-level:TRACE /sspi-module:/Users/user/Documents/projects/sspi-rs/target/debug/libsspi.dylib /auth-pkg-list:!ntlm,kerberos /tls:secrets-file:tls-secrets.txt 
```

image

## RD Gateway logon (Kerberos): RPC

```bash
./sdl-freerdp /v:DESKTOP-QELPR32.tbt.com /gateway:g:dgateway.tbt.com,type:rpc,u:t2,d:TBT,p:qqqQQQ111!!! /u:t2 /d:TBT /cert:ignore /p:qqqQQQ111!!! /log-level:TRACE /sspi-module:/Users/user/Documents/projects/sspi-rs/target/debug/libsspi.dylib /auth-pkg-list:!ntlm,kerberos /tls:secrets-file:tls-secrets.txt
```

image

## RD Gateway logon (NTLM)

```bash
./sdl-freerdp /v:DESKTOP-QELPR32.tbt.com /gateway:g:dgateway.tbt.com,u:t2,d:TBT,p:qqqQQQ111!!! /u:t2 /d:TBT /cert:ignore /p:qqqQQQ111!!! /log-level:TRACE /sspi-module:/Users/user/Documents/projects/sspi-rs/target/debug/libsspi.dylib /auth-pkg-list:ntlm,!kerberos /tls:secrets-file:tls-secrets.txt
```

image

PR #745, opened 2026-09-10
refactor(ffi) move SSPI type definitions to the `ffi-types`

TheBestTvarynka:
Hi,
I moved all FFI-related SSPI type definitions into the `ffy-types` crate. This cleaned up the code and lets us reuse FFI-types later.

Basically, I did the same as we did with WinSCard-related types back then.


CBenoit (approved):
Thank you! LGTM

PR #747, opened 2026-09-15
refactor: enable `indexing_slicing` lint

TheBestTvarynka:
Hi,
I didn't plan to work on lints, but I noticed the `sspi-rs` CI is still failing. The `release-pls` did not bump the `winscard` crate version: https://github.com/Devolutions/sspi-rs/actions/runs/33776962789/job/100720965324#step:4:135

```
 2026-09-03T16:10:15.426372Z  INFO winscard 0.3.3: already published
```

I was thinking of a small improvement in `winscard` to trigger `release-pls` and release a new version of `winscard`. I remembered that we have a lot of lints to enable.

This PR enables the `indexing_slicing` lint: https://rust-lang.github.io/rust-clippy/master/#indexing_slicing

> Checks for usage of indexing or slicing that may panic at runtime.
>
> This lint does not report on indexing or slicing operations that always panic, [out_of_bounds_indexing](https://rust-lang.github.io/rust-clippy/master/?search=indexing_slicing#out_of_bounds_indexing) already handles those cases.


TheBestTvarynka:
😅 Turned out that the publication problem is already fixed: https://github.com/Devolutions/sspi-rs/pull/746.

But that's okay. At least we have a new lint enabled 😄

crates/dpapi/src/crypto/mod.rs
```diff
@@ -290,10 +290,19 @@ fn kdf_concat(
...
+        .copy_from_slice(&key_guid.to_bytes_le());
+    buf.get_mut(16..20)
+        .expect("buf is 28 bytes long")
+        .copy_from_slice(&l0.to_le_bytes());
+    buf.get_mut(20..24)
+        .expect("buf is 28 bytes long")
+        .copy_from_slice(&l1.to_le_bytes());
+    buf.get_mut(24..28)
+        .expect("buf is 28 bytes long")
+        .copy_from_slice(&l2.to_le_bytes());
```
CBenoit: I think it may be worth using `#[expect(clippy::indexing_slicing, reason = "never indexed past 28")]` annotations well placed to avoid code noise when it’s not necessary

crates/dpapi-pdu/src/rpc/epm.rs
```diff
@@ -499,10 +499,12 @@ impl DecodeOwned for EntryHandle {
...
 
+        let (buf, uuid) = entry_handle_buf.split_at(4);
+
         Ok(if entry_handle_buf != Self::EMPTY_ENTRY_HANDLE {
             Self(Some((
-                u32::from_le_bytes(entry_handle_buf[0..4].try_into().unwrap()),
-                decode_uuid(&mut ReadCursor::new(&entry_handle_buf[4..]))?,
+                u32::from_le_bytes(buf.try_into().expect("4-byte buffer")),
+                decode_uuid(&mut ReadCursor::new(uuid))?,
             )))
```
CBenoit: praise: For instance, I think this code is better as it is rewritten here!

CBenoit (commented):
I think this lint is useful as a temporary audit tool, but I’m not sure we truly benefit from the extra code noise everywhere. My suggestion would be to:

- Run the lint once as an audit,
- Fix cases where the index or range is ultimately derived from:
  - a received token,
  - a decoded offset or length;
  - caller-controlled buffers;
  - ASN.1/DER or protocol metadata…
  - that is: adversarial input
- Leave clear local invariants alone, except if you can make the code _better_ by rewriting a non-panicking version.
- Optionally, enable in some narrow module with sensitive indexing, decoding code, especially against adverse input, do not enable workspace-wide. Use `#[expect]` with a reason to suppress the warnings when it’s not relevant.

Then, although it’s not static, fuzzing and property-test infrastructure should do the rest IMO, we need to fuzz anyway.

crates/dpapi-pdu/src/rpc/epm.rs
CBenoit: Although maybe, you could also get rid of the `try_into()` using https://doc.rust-lang.org/std/primitive.slice.html#method.split_first_chunk

